PR #4, opened 2019-10-13
UpdateSchema api for addition, updates and deletes

rdsr:
API for schema evolution. 

Added a core api under `UpdateSchema` .  The change is implemented using two visitors

First visitor traverses the updated schema and checks for additions and updates. In updates it also checks for renames, type widening, and documentation updates

Second visitor traverses the existing schema and checks for deletes in the updated schema.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/UpdateSchema.java
+++ b/api/src/main/java/org/apache/iceberg/UpdateSchema.java
@@ -197,7 +197,7 @@ public interface UpdateSchema extends PendingUpdate<Schema> {
   UpdateSchema deleteColumn(String name);
 
   /**
-   * Applies all the additions, updates [renames, type widening, field documentation] and deletes
+   * Applies all the additions and updates [type widening, field documentation]
    * from the input schema
    *
    * @param newSchema - Input schema from which updates are applied
--- a/api/src/main/java/org/apache/iceberg/UpdateSchema.java
+++ b/api/src/main/java/org/apache/iceberg/UpdateSchema.java
@@ -196,4 +196,12 @@ public interface UpdateSchema extends PendingUpdate<Schema> {
    */
   UpdateSchema deleteColumn(String name);
 
+  /**
+   * Applies all the additions, updates [renames, type widening, field documentation] and deletes
+   * from the input schema
+   *
+   * @param newSchema - Input schema from which updates are applied
+   * @return this for method chaining
+   */
+  UpdateSchema updateSchema(Schema newSchema);
 }
--- a/core/src/main/java/org/apache/iceberg/SchemaUpdate.java
+++ b/core/src/main/java/org/apache/iceberg/SchemaUpdate.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg;
 
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/apache/iceberg/UpdateSchema.java
```diff
@@ -196,4 +196,12 @@ default UpdateSchema updateColumn(String name, Type.PrimitiveType newType, Strin
    */
   UpdateSchema deleteColumn(String name);
 
+  /**
+   * Applies all the additions, updates [renames, type widening, field documentation] and deletes
+   * from the input schema
```
shardulm94: Should we call out the limitations of this API here? I don't think it is possible to support renames (since we won't know if it is delete+add instead). Deletion of a single column can probably be supported, but that combined with an addition cannot (because maybe it is a rename). So we should probably restrict ourselves to addition of field + type promotion.

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -192,6 +193,92 @@ public UpdateSchema updateColumnDoc(String name, String doc) {
     return this;
   }
 
+  @Override
+  public UpdateSchema updateSchema(Schema newSchema) {
+    TypeUtil.visit(newSchema, new ApplyUpdates(schema, this));
```
shardulm94: Should we reassign field ids here based on the old schema? ApplyUpdates requires field ids to match the old schema for correct behaviour.

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -192,6 +193,92 @@ public UpdateSchema updateColumnDoc(String name, String doc) {
...
+          if (!field.name().equals(newField.name())) {
+            updateSchema.renameColumn(name, newField.name());
+          }
+        }
+      }
+      return null;
+    }
+  }
+
+  private static class ApplyDeletes extends TypeUtil.SchemaVisitor<Void> {
```
shardulm94: Based on our discussion we should probably remove this for now.

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
shardulm94: Any precondition checks regarding the unsupported operations can also go here.

api/src/main/java/org/apache/iceberg/UpdateSchema.java
shardulm94: We may also say that the changes are all updates or all deletes, no mix and match. That should also be fine.

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -192,6 +193,92 @@ public UpdateSchema updateColumnDoc(String name, String doc) {
...
+          // updates
+          String name = join(parents, field.name());
+          if (field.type().isPrimitiveType() && !field.type().equals(newField.type())) {
+            updateSchema.updateColumn(name, newField.type().asPrimitiveType());
+          }
+          if (newField.doc() != null && !newField.doc().equals(field.doc())) {
+            updateSchema.updateColumnDoc(name, newField.doc());
+          }
+          if (!field.name().equals(newField.name())) {
+            updateSchema.renameColumn(name, newField.name());
```
shardulm94: Remove?

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -192,6 +193,92 @@ public UpdateSchema updateColumnDoc(String name, String doc) {
...
+    public Void struct(Types.StructType struct, List<Void> fieldResults) {
+      for (Types.NestedField newField : struct.fields()) {
+        Types.NestedField field = schema.findField(newField.fieldId());
+        String parents = DOT.join(fieldNames());
+        if (field == null) {
+          String name = join(parents, newField.name());
+          // found new field
+          if (parents.isEmpty()) {
+            // top level field
+            updateSchema.addColumn(name, newField.type());
```
shardulm94: We should still use the explicit parent parameter here since this will fail if the newField.name() itself contains `.`

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
```diff
@@ -192,6 +193,92 @@ public UpdateSchema updateColumnDoc(String name, String doc) {
...
+            // parent struct present
+            updateSchema.addColumn(parents, newField.name(), newField.type());
+          }
+          // else parent struct not present, so we will
+          // backtrack until we find a parent or reach top level
+        } else {
+          // updates
+          String name = join(parents, field.name());
+          if (field.type().isPrimitiveType() && !field.type().equals(newField.type())) {
+            updateSchema.updateColumn(name, newField.type().asPrimitiveType());
```
shardulm94: If the new field type is not primitive, it will lead to unfriendly error message. We should probably provide a better error message here.

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
rdsr: I think, by the time schema reaches this point it should already be assigned correct ids. Lets see how Spark does it. We can then revisit thi

api/src/main/java/org/apache/iceberg/UpdateSchema.java
rdsr: Lets address this upstream first

shardulm94: Update comment to remove renames and deletes for now?

core/src/main/java/org/apache/iceberg/SchemaUpdate.java
shardulm94: Sure. We can take a look later.

PR #9, opened 2019-10-22
LIHADOOP-49587: [Writer schema evolution] Update ReassignIds to assign ids to an incoming schema as per the table schema

rdsr:


api/src/main/java/org/apache/iceberg/types/ReassignIds.java
```diff
@@ -49,39 +54,53 @@ public Type struct(Types.StructType struct, Iterable<Type> fieldTypes) {
...
-        newFields.add(Types.NestedField.optional(sourceFieldId, field.name(), types.get(i), field.doc()));
+        newFields.add(Types.NestedField.optional(fieldId, field.name(), types.get(i), field.doc()));
       }
     }
 
     return Types.StructType.of(newFields);
   }
 
+  private int fieldId(Types.NestedField sourceField) {
+    if (sourceField == null || newField) {
```
shardulm94: What is the importance of the newField check? Is it not safe to assume that when `sourceField == null` it is a new field?

api/src/main/java/org/apache/iceberg/types/ReassignIds.java
```diff
@@ -49,39 +54,53 @@ public Type struct(Types.StructType struct, Iterable<Type> fieldTypes) {
...
+
     List<Types.NestedField> fields = struct.fields();
     int length = fields.size();
 
     List<Type> types = Lists.newArrayList(fieldTypes);
     List<Types.NestedField> newFields = Lists.newArrayListWithExpectedSize(length);
     for (int i = 0; i < length; i += 1) {
       Types.NestedField field = fields.get(i);
-      int sourceFieldId = sourceStruct.field(field.name()).fieldId();
+      Types.NestedField sourceField = sourceStruct.field(field.name());
```
shardulm94: Is it possible that `sourceStruct == null`? e.g. if this is a field within a new struct being added?

shardulm94: Ok... seems no because `sourceType = field.type();` is being set at line 92

api/src/main/java/org/apache/iceberg/types/ReassignIds.java
```diff
@@ -49,39 +54,53 @@ public Type struct(Types.StructType struct, Iterable<Type> fieldTypes) {
...
     }
-
-    this.sourceType = sourceField.type();
     try {
       return future.get();
     } finally {
       sourceType = sourceStruct;
+      if (!withinNewField) {
+        newField = false;
+      }
```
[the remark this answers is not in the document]
rdsr: Fixed

PR #11, opened 2019-10-25
Infer mapping from expected schema if file schema does not have ids and name mapping is missing

rdsr:


rdsr:
Discarding this in favor of upstreams fixes

PR #13, opened 2019-11-04
Added test artifacts to Iceberg-runtime module

rdsr:


rdsr:
discussed with @shardulm94 that this will bring in transitives from test jars

PR #19, opened 2020-02-05
Incremental scan followups

rdsr:


rdsr:
All changes were pulled in from Upstream for this specific patch

PR #21, opened 2020-02-06
Fail early with a proper exception if the metadata file is not existing

rdsr:


hive/src/main/java/org/apache/iceberg/hive/CustomHiveTableOperations.java
```diff
@@ -90,6 +90,11 @@ protected void doRefresh() {
           String errMsg = String.format("%s.%s is missing %s property", database, tableName, METADATA_LOCATION_PROP);
           throw new IllegalArgumentException(errMsg);
         }
+        if (!io().newInputFile(metadataLocation).exists()) {
+          String errMsg = String.format("%s.%s has non existent %s property",
```
shardulm94: The error message seems to not match the error here. Error is that the property exists but the file described in the property does not exists. Error message seems to indicate that the property is not set on the table. Can we throw a better error message? Preferably should have the metadata location included somewhere in the error message.

hive/src/main/java/org/apache/iceberg/hive/CustomHiveTableOperations.java
```diff
@@ -90,6 +90,11 @@ protected void doRefresh() {
           String errMsg = String.format("%s.%s is missing %s property", database, tableName, METADATA_LOCATION_PROP);
           throw new IllegalArgumentException(errMsg);
         }
+        if (!io().newInputFile(metadataLocation).exists()) {
```
rdsr: This is one more extra check we have to do. for every refresh. Once we are in stable state we should get rid of this.

rdsr:
Pushing this now. Will address followups if any

PR #22, opened 2020-02-24
Simplify CustomHiveCatalog

shardulm94:
- Rename CustomHiveCatalog to HiveMetadataPreservingCatalog to be more descriptive of the catalog behaviour
- Change access of a few variables in HiveCatalog which are required by the custom catalog to reduce code dup
- Change access of a few methods in HiveTableOperations which are required by the custom table ops to reduce code dup
- Update custom table ops code with changes from upstream
- Make Linkedin-specific changes more explicit by keeping the original lines but commenting them out


rdsr:
Minor check style issues. Once fixed I'll merge this. Thks!

PR #24, opened 2020-03-19
Support for non string partition columns

rdsr:


hive/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToIcebergType.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.hadoop.hive.serde2.typeinfo.DecimalTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.ListTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.MapTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.StructTypeInfo;
+import org.apache.iceberg.types.Type;
+import org.apache.iceberg.types.Types;
+
+
+public class HiveTypeToIcebergType extends HiveTypeUtil.HiveSchemaVisitor<Type> {
```
rdsr: We only using primitive conversion for now. But I thought we'd keep these classes since we can use these to convert schemas for tables which do not have avro.schema.literal

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableOperations.java
```diff
@@ -109,13 +109,14 @@ protected void doRefresh() {
     if (current().spec().fields().isEmpty()) {
       matchingDirectories = ImmutableList.of(getDirectoryInfo());
     } else {
-      matchingDirectories = getDirectoryInfosByFilter(expression);
+      matchingDirectories = getDirectoryInfosByFilter(expression, current().spec());
     }
 
-    Iterable<Iterable<DataFile>> filesPerDirectory = Iterables.transform(matchingDirectories, directory -> {
```
rdsr: this was getting hard for me to read, hence had to format these methods a little. Here and other places

hive/src/test/java/org/apache/iceberg/hive/legacy/TestLegacyHiveTableScan.java
```diff
@@ -93,44 +97,53 @@ public void testHiveScanUnpartitioned() throws Exception {
...
-    addPartition(table, ImmutableList.of("1"), AVRO, "A");
-    addPartition(table, ImmutableList.of("2"), AVRO, "B");
-    filesMatch(ImmutableMap.of("pcol=2/B", AVRO), hiveScan(table, Expressions.equal("pcol", "2")));
+    addPartition(table, ImmutableList.of("ds", 1), AVRO, "A");
+    addPartition(table, ImmutableList.of("ds", 2), AVRO, "B");
+    filesMatch(
+        ImmutableMap.of("pcol=ds/pIntCol=1/A", AVRO, "pcol=ds/pIntCol=2/B", AVRO),
+        hiveScan(table, Expressions.equal("pcol", "ds")));
+
+    AssertHelpers.assertThrows(
```
rdsr: Non string filtering is only supported for direct sql and direct sql only works for mysql backend.

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableOperations.java
```diff
@@ -139,7 +140,7 @@ private DirectoryInfo getDirectoryInfo() {
     }
   }
 
-  private List<DirectoryInfo> getDirectoryInfosByFilter(Expression expression) {
+  private List<DirectoryInfo> getDirectoryInfosByFilter(Expression expression, PartitionSpec spec) {
```
shardulm94: Any reason to pass the spec here instead of just using current().spec()?

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -50,33 +56,48 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
-  }
+    Types.StructType dataStructType = schema.asStruct();
+    List<Types.NestedField> fields = Lists.newArrayList(dataStructType.fields());
 
-  private static Schema addPartitionColumnsIfRequired(Schema schema, List<String> partitionColumns) {
-    List<Types.NestedField> fields = new ArrayList<>(schema.columns());
-    AtomicInteger fieldId = new AtomicInteger(10000);
-    partitionColumns.stream().forEachOrdered(column -> {
-      Types.NestedField field = schema.findField(column);
-      if (field == null) {
```
shardulm94: There are tables at LI which have datepartition in both data columns and partition columns, hence I had this check to only add the columns to the schema if they don't already have the partition column in the schema. Can we keep this check in the new logic?

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -50,33 +56,48 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
-            "Tables with non-string partition columns not supported yet");
+    Schema partitionSchema = partitionSchema(table.getPartitionKeys());
+    Types.StructType partitionStructType = partitionSchema.asStruct();
+    final Integer maxId = Collections.max(TypeUtil.indexById(dataStructType).keySet());
+    final Type partitionType = TypeUtil.assignFreshIds(partitionStructType, new TypeUtil.NextID() {
+      private int id = maxId;
+
+      @Override
+      public int get() {
+        id += 1;
```
shardulm94: Can we just add the partition fields to the schema and let TypeUtil.assignIncreasingFreshIds reassign all ids so we don't have to manually calculate MaxId or define our own NextID?

hive/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToIcebergType.java
```diff
@@ -0,0 +1,91 @@
...
+      case STRING:
+        return Types.StringType.get();
+      case BINARY:
+        return Types.BinaryType.get();
+      case BYTE:
+      case DECIMAL:
+        DecimalTypeInfo decimalTypeInfo = (DecimalTypeInfo) primitive;
+        return Types.DecimalType.of(decimalTypeInfo.precision(), decimalTypeInfo.scale());
+      case TIMESTAMP:
+      case DATE:
```
shardulm94: Can we just return date and timestamp types instead?

hive/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToIcebergType.java
```diff
@@ -0,0 +1,91 @@
...
+        return Types.IntegerType.get();
+      case LONG:
+        return Types.LongType.get();
+      case CHAR:
+      case VARCHAR:
+      case STRING:
+        return Types.StringType.get();
+      case BINARY:
+        return Types.BinaryType.get();
+      case BYTE:
```
shardulm94: case BYTE should not map to decimal, seems like a mistake

hive/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToIcebergType.java
```diff
@@ -0,0 +1,91 @@
...
+      case DECIMAL:
+        DecimalTypeInfo decimalTypeInfo = (DecimalTypeInfo) primitive;
+        return Types.DecimalType.of(decimalTypeInfo.precision(), decimalTypeInfo.scale());
+      case TIMESTAMP:
+      case DATE:
+      default:
+        throw new UnsupportedOperationException("UnSupported primitive type " + primitive);
```
shardulm94: Nit: Unsupported, s should be lowercase

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -50,33 +56,48 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
+        .stream()
+        .map(f -> Types.NestedField.optional(
+            fieldId.incrementAndGet(), f.getName(), icebergType(f.getType()), f.getComment()))
+        .collect(Collectors.toList());
+    return new Schema(partitionFields);
+  }
+
+  private static Type icebergType(String hiveTypeString) {
+    PrimitiveTypeInfo primitiveTypeInfo = TypeInfoFactory.getPrimitiveTypeInfo(hiveTypeString);
+    return new HiveTypeToIcebergType().primitive(primitiveTypeInfo);
```
shardulm94: Call HiveTypeUtil.visit instead? Since that is the standard way to call these visitors.

hive/src/test/java/org/apache/iceberg/hive/legacy/TestLegacyHiveTableScan.java
shardulm94: Can you split this failure into a different test please?

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableOperations.java
rdsr: Fixed

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
rdsr: I added this so that there's no conflict with existing partition and data schema for field ids. Since we are starting partition field ids from 10K I don't think we'll see any conflict. I'll remove

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
rdsr: Should we throw an exception instead? There could be some weird cases like
* What if the fields with the same name are of different types?
* What if the type is the same but have different values? 2020323 vs 2020-03-23 or something with completely different semantics?

hive/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToIcebergType.java
rdsr: I''l no map byte to anything for now and throw an exception.

shardulm94: Bytes are mapped to ints when going from Hive to Avro, or even when going from ORC to Iceberg. What do you think about this?

hive/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
shardulm94: I did a quick query and seems like only 25 user tables and 1 metrics table are affected. So I am okay throwing an exception.

rdsr: Thanks for the analysis.

hive/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToIcebergType.java
rdsr: Makes sense!

PR #26, opened 2020-04-07
Make LegacyHiveTable extensible.

yukuai518:
Change LegacyHiveTable constructor to be protected to allow inheritance.


yukuai518:
@shardulm94 Please review

PR #27, opened 2020-04-12
Add row filtering extension.

yukuai518:
Prototyped code, subject to change. Open the PR in order to discuss the potential way to refactor this.


api/src/main/java/org/apache/iceberg/RowFilter.java
```diff
@@ -0,0 +1,56 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public interface RowFilter<S, D> extends Closeable {
+  RowFilter DEFAULT = new RowFilter() {
```
rdsr, line 26: better to make it a class instead of anonymous

api/src/main/java/org/apache/iceberg/RowFilter.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public boolean accept(Object row) {
+      return true;
+    }
+  };
+
+  /**
+   * Expand the columns if required for the row filtering.
```
rdsr, line 43: please add examples on when this is required

api/src/main/java/org/apache/iceberg/RowFilter.java
```diff
@@ -0,0 +1,56 @@
...
+   * schema at the end.
+   *
+   * To make the conversion efficient, the expansion should
+   * be backward compatible, which means this method should
+   * append columns to the end of the original schema, so
+   * that reader can still use the original schema to read
+   * data.
+   */
+  S expandSchema(S originSchema);
+  boolean accept(D row);
```
rdsr, line 55: document

api/src/main/java/org/apache/iceberg/RowFilter.java
```diff
@@ -0,0 +1,56 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public interface RowFilter<S, D> extends Closeable {
```
rdsr, line 25: please document the class

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
```diff
@@ -180,6 +180,11 @@ public FileScanTask next() {
       this.fileScanTask = fileScanTask;
     }
 
+    @Override
+    public RowFilter createRowFilter(FileFormat format) {
```
rdsr, line 184: Does Predicate function changes per file?

rdsr: Why don't we instantiate the predicate once instead of per file?

orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
```diff
@@ -41,29 +41,31 @@
...
     this.config = config;
+    this.rowFilter = rowFilter;
   }
 
   @SuppressWarnings("unchecked")
   @Override
   public Iterator<T> iterator() {
     Reader orcFileReader = ORC.newFileReader(file, config);
     addCloseable(orcFileReader);
+    addCloseable(rowFilter);
```
rdsr, line 64: why should a filter be closable, isn't it stateless

core/src/main/java/org/apache/iceberg/BaseFileScanTask.java
yukuai518: Yes, the actual row filter is different between files. Each file will contain different segments (opal concepts), and each segment will have its own bitmap.

orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
yukuai518: Actually our row filter is based on the bitmap, which needs to be cleaned up when the task finishes.

shardulm94:
Incorporated into https://github.com/linkedin/iceberg/pull/28

PR #28, opened 2020-05-19
Row level filtering: Allow table scans to pass a row level filter for ORC files 

shardulm94:
Usage:
`TableScan`s which need row level filtering will implement `OrcRowFilterableFileScanTask` for the `FileScanTask`s which they return in `planTasks()`. Spark's `RowDataReader` will check if the task it wants to read implements this interface and provides the desired `OrcRowFilter` to ORC file readers.

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcRowFilter.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcRowFilter.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.orc;
+
+import java.io.Serializable;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.types.Type;
+
+/**
+ * [LINKEDIN ONLY]
+ * Represents a filter which can be applied to every row
+ * <p>
+ * Currently Row filter evaluation is only supported in Spark on unpartitioned tables for top level primitives
+ */
+public interface OrcRowFilter extends Serializable {
+
+  /**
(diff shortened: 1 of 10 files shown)


shardulm94:
FYI @yukuai518

spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
```diff
@@ -252,4 +258,29 @@ private static Object convertConstant(Type type, Object value) {
...
+    if (fileScanTask instanceof OrcRowFilterableFileScanTask) {
+      return ((OrcRowFilterableFileScanTask) fileScanTask).orcRowFilter();
+    }
+    return null;
+  }
+
+  private void validateRowFilterRequirements(FileScanTask task, OrcRowFilter filter) {
+    Preconditions.checkArgument(task.file().format() == FileFormat.ORC, "Row filter can only be applied to ORC files");
+    Preconditions.checkArgument(task.spec().fields().size() == 0,
+        "Row filter can only be applied to unpartitioned tables");
```
rdsr, line 278: I don't follow this requirement

orc/src/main/java/org/apache/iceberg/orc/OrcRowFilter.java
```diff
@@ -0,0 +1,47 @@
...
+   */
+  Schema requiredSchema();
+
+  /**
+   * Decides whether the row described by the provided field values should be kept or not.
+   * <p>
+   * The order of field values inside the values array is the same as the order of field ids returned by
+   * {@link #requiredSchema()}. The java class of the field value is dictated by {@link Type.TypeID#javaClass()}.
+   */
+  boolean shouldKeep(Object[] values);
```
rdsr, line 46: are these only top level columns. Should we call out as such?

spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
shardulm94: We don't handle it for unpartitioned tables today, else I will need to do special handling in RowDataReader. Since Opal tables are not unpartitioned, it isn't worth the effort.

orc/src/main/java/org/apache/iceberg/orc/OrcRowFilter.java
shardulm94: It is called out in the Javadoc of ORC row filter class.

spark/src/main/java/org/apache/iceberg/spark/source/RowDataReader.java
shardulm94: Actually you are correct. It was a limitation but I changed way I handle row level filtering and did not revisit this requirement. I will revisit this and remove if required.

shardulm94: I revisited this, the current process won't handle row filters on identity partitioned columns which come from metadata rather than data, since the values of these columns are not available while reading the data in OrcIterable. Since opal tables are not partitioned, I prefer keeping this restriction for now and revisit later if required.

rdsr: makes sense.

PR #29, opened 2020-05-22
Hack to work around #961 upstream 

shardulm94:
https://github.com/apache/incubator-iceberg/issues/961

Pretty benign hack


orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -331,11 +331,14 @@ private static TypeDescription buildOrcProjection(Integer fieldId, Type type, bo
...
-          }
-
+          // HACK: Disable the isRequired check to work around https://github.com/apache/incubator-iceberg/issues/961
+          // What we lose is a check; the file may not have the required column and so theoretically we may get null
+          // values for that column instead of failing. In reality though, if backward compatibility rules are followed,
+          // a new required field should never be added.
+          // if (isRequired) {
+          //   throw new IllegalArgumentException(
+          //       String.format("Field %d of type %s is required and was not found.", fieldId, type.toString()));
+          // }
```
wmoustafa, line 341: Is it possible to check that the parent isOptional and the child isRequired and pass those cases only? That would restrict the cases where the hack is in effect to more extreme ones.

shardulm94:
Superseeded by https://github.com/apache/iceberg/pull/1071

PR #37, opened 2020-09-30
Hive Metadata Scan: Do not use table snapshot summary for estimating statistics

shardulm94:
Iceberg uses a fast path to estimate stats using snapshot summary instead of iterating through file stats when not filters are present. This fails for our Hive table scan as it does not support snapshots.

For Hive table scans we should fallback to using the slower method which adds file sizes from underlying files.


HotSushi:
LGTM

PR #40, opened 2020-10-12
Serialize schema in InputSplit and pull upstream changes in #1557, 1564

HotSushi:
Hive queries that spawn MR jobs will work after this change because:
      1) Table schema in RecordReader is now read from the inputsplit 
      2) HiveSerde on mappers would use input config

Also added a test class which simulates linkedin's way of storing metadata and tests testcases written upstream such as testScanEmptyTable(), testScanTable(), testJoinTables()

cc: @shardulm94


HotSushi:
With changes in https://github.com/linkedin/iceberg/pull/43 : IcebergRecordReader can get correct schema

With changes in https://github.com/linkedin/iceberg/pull/45: HiveIcebergSerde can get correct schema

PR #48, opened 2020-11-20
Spark: Allow reading timestamp without time zone

shardulm94:
Spark only supports reading timestamp with time zone. However we have a lot of Hive tables which store timestamp without time zone.

In this PR, we modify the Spark code to allow reading timestamp without time zone as timestamp with time zone. Generally, this is not safe as timestamp without time zone is supposed to represent wall clock time semantics, i.e. no matter the reader/writer timezone 3PM should always be read as 3PM, but timestamp with time zone represents instant semantics, i.e the timestamp is adjusted so that the corresponding time in the reader timezone is displayed. However, at LinkedIn, all readers and writers are in the UTC timezone as our production machines are set to UTC. So, timestamp with/without time zone is the same. 

We put this feature behind a flag to not do this conversion by default and we will enable this flag at LinkedIn

cc: @HotSushi @wmoustafa

### DIFF
--- a/spark/src/main/java/org/apache/iceberg/spark/PruneColumnsWithoutReordering.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/PruneColumnsWithoutReordering.java
@@ -202,11 +202,6 @@ public class PruneColumnsWithoutReordering extends TypeUtil.CustomOrderSchemaVis
             "Cannot project decimal with incompatible precision: %s < %s",
             requestedDecimal.precision(), decimal.precision());
         break;
-      case TIMESTAMP:
-        Types.TimestampType timestamp = (Types.TimestampType) primitive;
-        Preconditions.checkArgument(timestamp.shouldAdjustToUTC(),
-            "Cannot project timestamp (without time zone) as timestamptz (with time zone)");
-        break;
       default:
     }
 
--- a/spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcReader.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcReader.java
@@ -104,6 +104,7 @@ public class SparkOrcReader implements OrcRowReader<InternalRow> {
           return OrcValueReaders.floats();
         case DOUBLE:
           return OrcValueReaders.doubles();
+        case TIMESTAMP:
         case TIMESTAMP_INSTANT:
           return SparkOrcValueReaders.timestampTzs();
         case DECIMAL:
--- a/spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
+++ b/spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
@@ -50,6 +50,7 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
@@ -151,6 +152,21 @@ public abstract class TestTimestampWithoutZone {
         read(unpartitioned.toString(), vectorized, "id", "ts"));
   }
 
+  @Rule
+  public ExpectedException exception = ExpectedException.none();
+
+  @Test
+  public void testUnpartitionedTimestampWithoutZoneError() {
+    exception.expect(IllegalArgumentException.class);
+    exception.expectMessage("Spark does not support timestamp without time zone fields");
+
+    spark.read().format("iceberg")
+        .option("vectorization-enabled", String.valueOf(vectorized))
+        .option("read-timestamp-without-zone", "false")
+        .load(unpartitioned.toString())
+        .collectAsList();
+  }
+
   private static Record projectFlat(Schema projection, Record record) {
     Record result = GenericRecord.create(projection);
     List<Types.NestedField> fields = projection.asStruct().fields();
@@ -163,12 +179,10 @@ public abstract class TestTimestampWithoutZone {
 
   public static void assertEqualsSafe(Types.StructType struct,
                                       List<Record> expected, List<Row> actual) {
-    // TODO: match records by ID
-    int numRecords = Math.min(expected.size(), actual.size());
-    for (int i = 0; i < numRecords; i += 1) {
+    Assert.assertEquals("Number of results should match expected", expected.size(), actual.size());
+    for (int i = 0; i < expected.size(); i += 1) {
       GenericsHelpers.assertEqualsSafe(struct, expected.get(i), actual.get(i));
     }
-    Assert.assertEquals("Number of results should match expected", expected.size(), actual.size());
   }
 
   private List<Record> testRecords(Schema schema) {
--- a/spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
+++ b/spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
@@ -1,0 +1,213 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark.source;
+
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Locale;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DataFiles;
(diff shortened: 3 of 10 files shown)


spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
```diff
@@ -0,0 +1,213 @@
...
+        record(schema, 8L, parseToLocal("2017-12-21T15:21:51.237521"), "global"),
+        record(schema, 9L, parseToLocal("2017-12-21T15:02:15.230570"), "goldfish")
+    );
+  }
+
+  private static List<Row> read(String table, boolean vectorized) {
+    return read(table, vectorized, "*");
+  }
+
+  private static List<Row> read(String table, boolean vectorized, String select0, String... selectN) {
```
HotSushi, line 193: Can we add a test case to assert that an exception is thrown if read-timestamp-without-zone is false?

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcReader.java
```diff
@@ -104,6 +104,7 @@ private ReadBuilder(Map<Integer, ?> idToConstant) {
           return OrcValueReaders.floats();
         case DOUBLE:
           return OrcValueReaders.doubles();
+        case TIMESTAMP:
         case TIMESTAMP_INSTANT:
           return SparkOrcValueReaders.timestampTzs();
```
HotSushi, line 109: Does SparkOrcValueReaders.timestampTzs() read both LocalDateTime and OffsetDateTime?

spark/src/main/java/org/apache/iceberg/spark/PruneColumnsWithoutReordering.java
```diff
@@ -202,11 +202,6 @@ public Type primitive(Type.PrimitiveType primitive) {
             "Cannot project decimal with incompatible precision: %s < %s",
             requestedDecimal.precision(), decimal.precision());
         break;
-      case TIMESTAMP:
-        Types.TimestampType timestamp = (Types.TimestampType) primitive;
-        Preconditions.checkArgument(timestamp.shouldAdjustToUTC(),
-            "Cannot project timestamp (without time zone) as timestamptz (with time zone)");
```
HotSushi, line 208: Was this support unavailable just for ORC? or for all readers?

If it's for all readers, was there no change similar to SparkORCReader or VectorizedSparkORCReader required in those respective readers?

shardulm94: Avro and Parquet do not support timestamp natively, both Ts and TsTz are just longs for Avro and Parquet.

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcReader.java
shardulm94: Spark uses longs (microseconds since epoch) to store timestamps, so SparkOrcValueReaders.timestampTzs() just needs to return longs.

spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  public static void assertEqualsSafe(Types.StructType struct,
+                                      List<Record> expected, List<Row> actual) {
+    // TODO: match records by ID
+    int numRecords = Math.min(expected.size(), actual.size());
+    for (int i = 0; i < numRecords; i += 1) {
+      GenericsHelpers.assertEqualsSafe(struct, expected.get(i), actual.get(i));
+    }
+    Assert.assertEquals("Number of results should match expected", expected.size(), actual.size());
```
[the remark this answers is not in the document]
shardulm94: Ack. Done.

spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
shardulm94: Ack Done

spark2/src/main/java/org/apache/iceberg/spark/source/Reader.java
```diff
@@ -166,6 +170,13 @@
     this.batchSize = options.get("batch-size").map(Integer::parseInt).orElseGet(() ->
         PropertyUtil.propertyAsInt(table.properties(),
           TableProperties.PARQUET_BATCH_SIZE, TableProperties.PARQUET_BATCH_SIZE_DEFAULT));
+    // Allow reading timestamp without time zone as timestamp with time zone. Generally, this is not safe as timestamp
+    // without time zone is supposed to represent wall clock time semantics, i.e. no matter the reader/writer timezone
+    // 3PM should always be read as 3PM, but timestamp with time zone represents instant semantics, i.e the timestamp
+    // is adjusted so that the corresponding time in the reader timezone is displayed. However, at LinkedIn, all readers
+    // and writers are in the UTC timezone as our production machines are set to UTC. So, timestamp with/without time
+    // zone is the same.
```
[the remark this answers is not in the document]
shardulm94: Done

wmoustafa:
I think what we should do is:
1- Check the list of functions in Hive that work with timestamps (e.g., second, hour, datediff, unix_timestamp), and verify if they take "Timestamp" or "String" as input. Based on the [Hive manual](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF), they take strings. This exercise should consider Hive 1.1 preferably.
2- Check the same list of functions in Spark, and see if Spark would deviate out of the box (due to Spark in some cases dealing with "Timestamp with Timezone). If Hive and Spark both expect Strings, is the string format a factor (if it contains timezone or if it does not)? Does it mean the query is not portable between Hive and Spark? (all this is research for Hive/Spark compatibility).
3- See if Dali introduces UTC, any of the Hive behavior in (1) or Spark behavior in (2) would break.

spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
```diff
@@ -0,0 +1,227 @@
...
+    this.vectorized = vectorized;
+  }
+
+  private File parent = null;
+  private File unpartitioned = null;
+  private List<Record> records = null;
+
+  @Before
+  public void writeUnpartitionedTable() throws IOException {
+    this.parent = temp.newFolder("TestFilteredScan");
```
kbendick, line 113: Happened to find myself here when @shardulm94 pointed out this PR in the Iceberg community sync, so feel free to ignore / resolve if this isn't a concern for you guys. 👍 

Possible nit: Looks like maybe this was copied from a `TestFilteredScan` test? Seems like the temp folder should be `TestTimestampWithoutZone`.

spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
```diff
@@ -0,0 +1,227 @@
...
+    this.format = format;
+    this.vectorized = vectorized;
+  }
+
+  private File parent = null;
+  private File unpartitioned = null;
+  private List<Record> records = null;
+
+  @Before
+  public void writeUnpartitionedTable() throws IOException {
```
wmoustafa, line 112: What is the significance of `Unpartitioned`?

spark2/src/main/java/org/apache/iceberg/spark/source/Reader.java
```diff
@@ -189,6 +202,8 @@ private Expression filterExpression() {
 
   private StructType lazyType() {
     if (type == null) {
+      Preconditions.checkArgument(readTimestampWithoutZone || !hasTimestampWithoutZone(lazySchema()),
+          "Spark does not support timestamp without time zone fields");
```
wmoustafa, line 206: If we do nothing here, would not the exception be thrown later?

spark2/src/main/java/org/apache/iceberg/spark/source/Reader.java
```diff
@@ -340,14 +355,22 @@ public boolean enableBatchRead() {
 
       boolean onlyPrimitives = lazySchema().columns().stream().allMatch(c -> c.type().isPrimitiveType());
 
+      boolean hasTimestampWithoutZone = hasTimestampWithoutZone(lazySchema());
+
       boolean hasNoDeleteFiles = tasks().stream().noneMatch(TableScanUtil::hasDeletes);
 
       this.readUsingBatch = batchReadsEnabled && hasNoDeleteFiles && ((allOrcFileScanTasks && hasNoRowFilters) ||
-          (allParquetFileScanTasks && atLeastOneColumn && onlyPrimitives));
+          (allParquetFileScanTasks && atLeastOneColumn && onlyPrimitives && !hasTimestampWithoutZone));
```
wmoustafa, line 363: Could you please explain why this is required now (especially no special treatment existed for schemas with Timestamp before this patch)?

spark2/src/main/java/org/apache/iceberg/spark/source/Reader.java
wmoustafa: Never mind. The exception is actually deleted from the later steps.

spark2/src/main/java/org/apache/iceberg/spark/source/Reader.java
shardulm94: I don't quite remember this because it was a while ago, but my guess is that vectorized reads in Parquet do not support timestamp without zone type. Parquet vectorized reads are only supported in Spark and so my guess is that the implementation did not bother supporting it.

spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
shardulm94: Nothing more than just saying that the table its writing is unpartitioned. This code was copied from another test which probably had two methods `writeUnpartitionedTable` and `writePartitionedTable`.

spark/src/test/java/org/apache/iceberg/spark/source/TestTimestampWithoutZone.java
shardulm94: Thanks @kbendick! Fixed this.

wmoustafa (approved):
LGTM.

PR #51, opened 2020-12-10
Fix pushdown of non-partition predicates within NOT

shardulm94:
`partitionCol = 1 AND NOT(nonpartitionCol = 2)` was getting simplified to `partitionCol = 1 AND NOT(TRUE)` and consequently `FALSE` when removing partition predicates.
Now the flow looks like `partitionCol = 1 AND NOT(nonpartitionCol = 2)` => `partitionCol = 1 AND nonpartitionCol != 2)` => `partitionCol = 1 AND REMOVED` => `partitionCol = 1`

Other expressions like  `NOT(partitionCol = 1 AND nonpartitionCol = 2)` will now translate to `NOT(partitionCol = 1) OR NOT(nonpartitionCol = 2)` => `partitionCol != 1 AND nonpartitionCol != 2` => `partitionCol != 1 AND REMOVED` => `partitionCol != 1`

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
@@ -33,6 +33,7 @@ import org.apache.iceberg.expressions.UnboundTerm;
 
 class HiveExpressions {
 
+  private static final Expression REMOVED = (Expression) () -> null;
   private HiveExpressions() {}
 
   /**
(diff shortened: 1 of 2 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
```diff
@@ -48,9 +49,17 @@ private HiveExpressions() {}
    */
   static Expression simplifyPartitionFilter(Expression expr, Set<String> partitionColumnNames) {
     try {
-      Expression partitionPredicatesOnly = ExpressionVisitors.visit(expr,
+      // Pushing down NOTs is critical for the correctness of RemoveNonPartitionPredicates
+      // Without pushdown NOT(P and NP) will be written to NOT(P)
```
wmoustafa, line 53: With simply ignoring NP, ..
Also may clarify what P and NP are

wmoustafa (approved):
Thanks @shardulm94! Nice patch.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
shardulm94: Done

PR #52, opened 2020-12-17
Hive Metadata Scan: Support case insensitive name mapping

shardulm94:
This PR intends to support cases where table schema does not exactly match the file schema in casing for our Hive tables.

- Adds a case insensitive mode to `NameMapping`
   -  Search for an exact match first and if not found searches for a case insensitive match
- Update `NameMappingParser` to serialize and deserialize the new case sensitivity field as JSON
- Add tests to display and verify behaviour of base and corner cases, comments in the test have explanations
- LegacyHiveCatalog will now provide case insensitive mapping for Hive tables
- Removed some old LI specific code in ORC which is no longer necessary after name mapping support upstream

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableMetadata.java
+++ b/core/src/main/java/org/apache/iceberg/TableMetadata.java
@@ -71,17 +71,6 @@ public class TableMetadata implements Serializable {
     return newTableMetadata(schema, spec, location, properties, DEFAULT_TABLE_FORMAT_VERSION);
   }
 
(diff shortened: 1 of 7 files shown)


wmoustafa (commented):
Looks good to me. Only thing we could consider in the future is whether using a case insensitive Map implementation (e.g., Apache's [CaseInsensitiveMap](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/CaseInsensitiveMap.html)) would simply the implementation, but possibly not by too much.

wmoustafa (approved):

PR #55, opened 2021-02-02
Hive Metastore: Merge hive and avro schema if not consistent

funcheetah:


hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,698 @@
...
+import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.hadoop.hive.serde2.avro.AvroSerdeUtils;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class LegacyHiveSchemaUtils {
```
shardulm94, line 42: The code in this class is way too verbose. This can be heavily simplified by using Visitors

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,272 @@
...
+import org.apache.hadoop.hive.serde2.typeinfo.MapTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.StructTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.UnionTypeInfo;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.codehaus.jackson.node.JsonNodeFactory;
+
+
+public class HiveTypeToAvroType {
```
shardulm94, line 42: Can we add test cases?

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,26 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
+      org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(schemaStr);
+      org.apache.avro.Schema hiveSchema = LegacyHiveSchemaUtils.convertHiveSchemaToAvro(table);
+      org.apache.avro.Schema avroSchemaWithoutNullable =
+          LegacyHiveSchemaUtils.extractActualTypeIfFieldIsNullableTypeRecord(avroSchema);
+      org.apache.avro.Schema hiveSchemaWithoutNullable =
+          LegacyHiveSchemaUtils.extractActualTypeIfFieldIsNullableTypeRecord(hiveSchema);
+
+      org.apache.avro.Schema tableSchema = avroSchema;
+      boolean isHiveSchemaEvolved =
+          LegacyHiveSchemaUtils.isRecordSchemaEvolved(avroSchemaWithoutNullable, hiveSchemaWithoutNullable);
```
shardulm94, line 68: Seems like isRecordSchemaEvolved has to traverse the whole Schema tree. What do we gain by checking this first rather than just merging directly?

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,543 @@
...
+    String recordNamespace = table.getDbName() + "." + recordName;
+
+    final List<FieldSchema> cols = new ArrayList<>();
+
+    cols.addAll(table.getSd().getCols());
+    if (isPartitioned(table)) {
+      cols.addAll(getPartitionCols(table));
+    }
+
+    return convertFieldSchemaToAvroSchema(recordName, recordNamespace, true, cols);
```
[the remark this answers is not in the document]
rzhang10: I see this is a tail reference/call, we can just move the code in `convertFieldSchemaToAvroSchema` inside `convertHiveSchemaToAvro`. But aside from that, what do you mean by "create a struct field schema"? I see the `parseSchemaFromStruct` method from `HiveTypeToAvroType`, but it is a private method, are you referring to this method?

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
rzhang10: will do later, but the integration test I ran already passed all the tables.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
rzhang10: I think this is actually a good point. I think the 2 logic can be combined into just one pass.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,253 @@
...
+    this.mkFieldsOptional = mkFieldsOptional;
+  }
+
+  Schema convertFieldsTypeInfoToAvroSchema(String recordNamespace, String recordName, List<String> fieldNames,
+      List<TypeInfo> fieldTypeInfos) {
+    final List<Schema.Field> fields = new ArrayList<>();
+    for (int i = 0; i < fieldNames.size(); ++i) {
+      final TypeInfo fieldTypeInfo = fieldTypeInfos.get(i);
+      String fieldName = fieldNames.get(i);
+      fieldName = removePrefix(fieldName);
```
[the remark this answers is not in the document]
rzhang10: Yeah, I think it can be removed? also since we are dealing with hive, I feel the field names won't contain `.` anyways.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,13 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
     String schemaStr = props.get("avro.schema.literal");
     Schema schema;
     if (schemaStr != null) {
-      schema = AvroSchemaUtil.toIceberg(new org.apache.avro.Schema.Parser().parse(schemaStr));
+      org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(schemaStr);
+      org.apache.avro.Schema hiveSchema = LegacyHiveSchemaUtils.convertHiveSchemaToAvro(table);
```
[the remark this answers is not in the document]
rzhang10: Good catch, I think I can change the signature of `convertHiveSchemaToAvro` to add this boolean flag parameter to `convertHiveSchemaToAvro(Table table, boolean mkFieldsOptional), so that this function directly return a non-null version of the schema.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,272 @@
...
+  private int recordCounter;
+  private final String namespace;
+
+  private final boolean mkFieldsOptional;
+
+  // Additional numeric type, similar to other logical type names in AvroSerde
+  private static final String SHORT_TYPE_NAME = "short";
+  private static final String BYTE_TYPE_NAME = "byte";
+
+  public HiveTypeToAvroType(String namespace, boolean mkFieldsOptional) {
```
wmoustafa, line 52: Does it make sense to convert this to a utility class and move those parameters to `convertFieldsTypeInfoToAvroSchema`?

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,272 @@
...
+    this.recordCounter = 0;
+    this.namespace = namespace;
+
+    this.mkFieldsOptional = mkFieldsOptional;
+  }
+
+  Schema convertFieldsTypeInfoToAvroSchema(String recordNamespace, String recordName, List<String> fieldNames,
+      List<TypeInfo> fieldTypeInfos) {
```
wmoustafa, line 60: Can we avoid using list of field names and list of field types throughout the PR? For example instead of of `List<String> fieldNames` and `List<TypeInfo> fieldTypeInfos`, we just pass `StructTypeInfo`. Usually the input is already a `StructTypeInfo` and then it is broken down to two lists then dealt with here. In cases where the original input comes as two lists, we can combine them using `TypeInfoFactory. getStructTypeInfo()` from Hive.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,272 @@
...
+  Schema convertTypeInfoToAvroSchema(TypeInfo typeInfo, String recordNamespace, String recordName) {
+    Schema schema;
+    ObjectInspector.Category category = typeInfo.getCategory();
+
+    switch (category) {
+      case STRUCT:
+        // We don't cache the structType because otherwise it could be possible that a field
+        // "lastname" is of type "firstname", where firstname is a compiled class.
+        // This will lead to ambiguity.
+        schema = parseSchemaFromStruct((StructTypeInfo) typeInfo, recordNamespace, recordName);
```
wmoustafa, line 91: Can we rename this and other methods to something like `convertStructTypeInfoToAvroSchema`?

wmoustafa (changes requested):
Still going through the patch. Currently, there are many ways conversions take place and seems they could be simplified. For example, we can go from Hive type string to Hive TypeInfo (one type info to represent the whole schema) then to Avro schema.

PR #56, opened 2021-02-18
merge hive native schema and avro schema literal if they are inconsistent

rzhang10:
add support to merge hive native schema and avro schema literal if they are not consistent, the merging logic is to use hive schema as the source of truth for types, while augmenting other metadata/attributes such as casing, docstring, default value etc from avro schema.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
@@ -39,18 +39,11 @@ import org.codehaus.jackson.node.JsonNodeFactory;
 public class HiveTypeToAvroType {
   private int recordCounter;
 
-  private final boolean mkFieldsOptional;
-
   // Additional numeric type, similar to other logical type names in AvroSerde
   private static final String SHORT_TYPE_NAME = "short";
   private static final String BYTE_TYPE_NAME = "byte";
 
-  public HiveTypeToAvroType(boolean mkFieldsOptional) {
-    this.recordCounter = 0;
-    this.mkFieldsOptional = mkFieldsOptional;
-  }
-
-  Schema convertTypeInfoToAvroSchema(TypeInfo typeInfo, String recordNamespace, String recordName) {
+  static Schema convertTypeInfoToAvroSchema(TypeInfo typeInfo, String recordNamespace, String recordName) {
     Schema schema;
     ObjectInspector.Category category = typeInfo.getCategory();
 
@@ -80,14 +73,14 @@ public class HiveTypeToAvroType {
     return schema;
   }
 
-  private Schema parseSchemaFromUnion(UnionTypeInfo unionTypeInfo, final String recordNamespace,
+  private static Schema parseSchemaFromUnion(UnionTypeInfo unionTypeInfo, final String recordNamespace,
                                       final String recordName) {
     List<TypeInfo> typeInfos = unionTypeInfo.getAllUnionObjectTypeInfos();
 
     // A union might contain duplicate struct typeinfos because the underlying Avro union has two Record types with
     // different names but the same internal structure.
     // In the case of duplicate typeinfos, we generate a new record type for each struct typeinfo.
-
+    int recordCounter = 0;
     List<Schema> schemas = new ArrayList<>();
 
     for (TypeInfo typeInfo : typeInfos) {
@@ -112,7 +105,7 @@ public class HiveTypeToAvroType {
 
   // Previously, Hive use recordType[N] as the recordName for each structType,
   // with the change we made in LIHADOOP-36761, the new record name will be in the form of "structNamespace.structName"
-  private Schema parseSchemaFromStruct(final StructTypeInfo typeInfo, final String recordNamespace,
+  private static Schema parseSchemaFromStruct(final StructTypeInfo typeInfo, final String recordNamespace,
       final String recordName) {
     final List<Schema.Field> avroFields = new ArrayList<>();
 
@@ -135,13 +128,14 @@ public class HiveTypeToAvroType {
     return recordSchema;
   }
 
-  private Schema parseSchemaFromList(final ListTypeInfo typeInfo, final String recordNamespace,
+  private static Schema parseSchemaFromList(final ListTypeInfo typeInfo, final String recordNamespace,
       final String recordName) {
     Schema listSchema = convertTypeInfoToAvroSchema(typeInfo.getListElementTypeInfo(), recordNamespace, recordName);
     return Schema.createArray(listSchema);
   }
 
-  private Schema parseSchemaFromMap(final MapTypeInfo typeInfo, final String recordNamespace, final String recordName) {
+  private static Schema parseSchemaFromMap(final MapTypeInfo typeInfo, final String recordNamespace,
+                                           final String recordName) {
     final TypeInfo keyTypeInfo = typeInfo.getMapKeyTypeInfo();
     final PrimitiveObjectInspector.PrimitiveCategory pc = ((PrimitiveTypeInfo) keyTypeInfo).getPrimitiveCategory();
     if (pc != PrimitiveObjectInspector.PrimitiveCategory.STRING) {
@@ -154,7 +148,7 @@ public class HiveTypeToAvroType {
     return Schema.createMap(valueSchema);
   }
 
-  private Schema parseSchemaFromPrimitive(PrimitiveTypeInfo primitiveTypeInfo) {
+  private static Schema parseSchemaFromPrimitive(PrimitiveTypeInfo primitiveTypeInfo) {
     Schema schema;
     switch (primitiveTypeInfo.getPrimitiveCategory()) {
       case LONG:
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
@@ -1,0 +1,227 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.hive.legacy;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.avro.Schema;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.hive.serde2.avro.AvroSerDe;
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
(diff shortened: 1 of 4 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+      Schema.Field evolvedMapField =
+          new Schema.Field(oldField.name(), evolvedMapSchema, oldField.doc(), oldField.defaultVal());
+
+      mergeAndAppendField(sourceField, evolvedMapField, fieldAssembler);
+    }
+  }
+
+  private static void handleArraySchemaMerge(Map<String, Queue<Schema.Field>> metaDataFieldsMap,
+      SchemaBuilder.FieldAssembler<Schema> fieldAssembler, Schema.Field sourceField, String sourceFieldName,
+      Schema sourceFieldSchema, String sourceFieldNameLowercase, boolean isNewlyAddedField) {
```
wmoustafa, line 334: I am not able to follow what is going on in this method, especially with the new names we keep introducing: evolved, old, etc. We have two schemas: hive and supplemental. Let us reuse the names whenever possible.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+
+    return partKeys;
+  }
+
+  private static void mergeAndAppendField(Schema.Field sourceField, Schema.Field metaDataField,
+      SchemaBuilder.FieldAssembler<Schema> fieldAssembler) {
+    Schema.Type actualMetaDataFieldType = extractActualTypeIfFieldIsNullableType(metaDataField.schema()).getType();
+    boolean isSameType = sourceField.schema().getType().equals(actualMetaDataFieldType);
+    boolean isMetaDataFieldEnumType = metaDataField.schema().getType().equals(Schema.Type.ENUM);
+    boolean useMetaDataType = isSameType || isMetaDataFieldEnumType;
```
wmoustafa, line 680: If our contract is to use the Hive schema types, we should not borrow the Enum types from the supplemental schema as well.

wmoustafa: What is the value of making a special case to use MetaDataType if is the same type? Why can we just use the Hive type by default, and then we would not have to make this condition altogether? (especially after removing `isMetaDataFieldEnumType` as well.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+
+    Object defaultValue = metaDataField.defaultVal();
+    if (defaultValue != null) {
+      fieldAssembler.name(metaDataField.name()).doc(metaDataField.doc())
+          .type(useMetaDataType ? metaDataField.schema() : sourceField.schema()).withDefault(defaultValue);
```
wmoustafa, line 685: Take `fieldAssembler.name(metaDataField.name()).doc(metaDataField.doc()).type(useMetaDataType ? metaDataField.schema() : sourceField.schema())` out of the condition and only adjust the default value inside the `if` statement. Is the default `noDefault()`? Then maybe no need for the `noDefault()` branch, but please double check.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+    List<FieldSchema> partKeys = tableOrView.getPartitionKeys();
+    if (partKeys == null) {
+      partKeys = new ArrayList<>();
+      tableOrView.setPartitionKeys(partKeys);
+    }
+
+    return partKeys;
+  }
+
+  private static void mergeAndAppendField(Schema.Field sourceField, Schema.Field metaDataField,
```
wmoustafa, line 675: `applySupplementalFieldToHiveField()`. Also keep the order of the parameters consistent.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+              "New Schema is RECORD type, while old Schema is UNION type without RECORD type in it");
+        }
+      }
+
+      throw new IllegalArgumentException(
+          "New Schema is RECORD type, while old Schema is nether RECORD type nor UNION type");
+    }
+  }
+
+  private static Schema mergeArrayAndMapFieldSchema(Schema oldFieldSchema, Schema newFieldSchema,
```
wmoustafa, line 392: This method is very complex. What is it trying to do? It also calls a few complex methods. Can you simplify it? We can setup a meeting to discuss.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,26 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
+      org.apache.avro.Schema tableSchema = avroSchema;
+      boolean isHiveSchemaEvolved =
+          LegacyHiveSchemaUtils.isRecordSchemaEvolved(avroSchemaWithoutNullable, hiveSchemaWithoutNullable);
+
+      if (isHiveSchemaEvolved) {
+        try {
+          tableSchema = LegacyHiveSchemaUtils.mergeRecordSchema(avroSchema, hiveSchemaWithoutNullable);
+        } catch (Exception e) {
+          tableSchema = avroSchema;
+        }
```
wmoustafa, line 75: This is weird. What is the reason we need to do this?

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,26 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
+      org.apache.avro.Schema avroSchema = new org.apache.avro.Schema.Parser().parse(schemaStr);
+      org.apache.avro.Schema hiveSchema = LegacyHiveSchemaUtils.convertHiveSchemaToAvro(table);
+      org.apache.avro.Schema avroSchemaWithoutNullable =
+          LegacyHiveSchemaUtils.extractActualTypeIfFieldIsNullableTypeRecord(avroSchema);
+      org.apache.avro.Schema hiveSchemaWithoutNullable =
+          LegacyHiveSchemaUtils.extractActualTypeIfFieldIsNullableTypeRecord(hiveSchema);
+
+      org.apache.avro.Schema tableSchema = avroSchema;
+      boolean isHiveSchemaEvolved =
+          LegacyHiveSchemaUtils.isRecordSchemaEvolved(avroSchemaWithoutNullable, hiveSchemaWithoutNullable);
```
wmoustafa, line 68: Why are we describing this in terms of `evolution`? We are simply comparing a Hive schema to a Supplemental schema. They either match or not match (according to some criteria that is defined in the method). Otherwise, there is no `evolution`, `new`, `old`, etc.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,26 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
...
+
+      if (isHiveSchemaEvolved) {
```
wmoustafa, line 70: I am not sure why this check is required. Could you explain why it matters? If it turns it is not needed, this patch will be significantly simplified.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+        // oldField is added in new schema
+        LOG.info("Schema is evolved. The field {} is added in new schema", newField.name());
+        return true;
+      }
+    }
+
+    for (Schema.Field oldField : oldSchemaFields) {
+      Schema.Field newField = newSchemaFieldsMap.get(oldField.name().toLowerCase());
+
+      if (isSchemaEvolved(oldField.schema(), newField.schema())) {
```
wmoustafa, line 587: I do not think there is a reason to used the word `evolved` in this patch.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveSchemaUtils.java
```diff
@@ -0,0 +1,691 @@
...
+    }
+
+    List<Schema> unionSchemas = new ArrayList<>();
+    unionSchemas.add(Schema.create(Schema.Type.NULL));
+    unionSchemas.add(schema);
+
+    return Schema.createUnion(unionSchemas);
+  }
+
+  static boolean isRecordSchemaEvolved(Schema oldSchema, Schema newSchema) {
```
wmoustafa, line 548: Why do we have two methods with slightly different names `isRecordSchemaEvolved` and `isSchemaEvolved`?

wmoustafa (changes requested):
Overall, seems there are lots of opportunities to simplify. In addition to regular comments, I have pointed out the places where the patch can be simplified. Let us simplify and see how it looks then.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,221 @@
...
+    return schema;
+  }
+
+  private static Schema parseSchemaFromUnion(UnionTypeInfo unionTypeInfo, final String recordNamespace,
+                                      final String recordName) {
+    List<TypeInfo> typeInfos = unionTypeInfo.getAllUnionObjectTypeInfos();
+
+    // A union might contain duplicate struct typeinfos because the underlying Avro union has two Record types with
+    // different names but the same internal structure.
+    // In the case of duplicate typeinfos, we generate a new record type for each struct typeinfo.
```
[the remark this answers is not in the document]
shardulm94: Different names refers to different Avro namespace+record name https://avro.apache.org/docs/current/spec.html#Unions, not a field name. 
e.g. a union might have two member field with are `struct<a:int,b:int>`, but if they have a different Avro namespace+record name, then it is valid.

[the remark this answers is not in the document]
wmoustafa: That makes sense. Thanks for the clarification.

shardulm94:
The bugs and differences I observed through the tests implemented in #57

Bugs
- `TestMergeHiveSchemaWithAvro#shouldRetainExtraFieldsFromHive` If there are extra top level struct fields in Hive, the nullability of those fields is not handled correctly. e.g if the extra field is a `struct<a:int>`, the `struct` is marked as optional, but `a` is marked as required.  `a` should also be nullable.
- `TestMergeHiveSchemaWithAvro#shouldHandleLists` If there are extra top level list or map fields in Hive, there are not retained in the resultant schema.
- `TestMergeHiveSchemaWithAvro#shouldSanitizeIncompatibleFieldNames` If the table contains field names not allowed by Avro identifier spec i.e. [A-Za-z0-9_], they are not sanitized and Hive to Avro conversion fails

Differences
- `TestMergeHiveSchemaWithAvro#shouldFailIfContainerTypesDontMatch` This implementation does not fail if Hive type is a container type viz. STRUCT, LIST and MAP and if they Avro type does not match.

We may need more comprehensive testing for deeper levels of nesting of list/map/structs. I didn't feel is was absolutely necessary in #57 because of the way the visitor works, but it might be helpful to add them.

PR #57, opened 2021-02-22
Alternative implementation of #55 Hive Metadata Scan: Merge Hive and Avro schemas if they are inconsistent

shardulm94:
This is an alternative implementation to #55 
This PR uses a SchemaWithPartner visitor to traverse a Hive schema along with a partner Avro schema to augment the Hive schema with field casing, nullability, default values, etc and get a resultant Avro schema.
- `MergeHiveSchemaWithAvro` contains the core merging logic
- `HiveSchemaWithPartnerVisitor` an abstract visitor containing the logic to traverse a Hive schema alongside a partner
- `HiveTypeToAvroType` converts Hive type to corresponding Avro type. We only use this to convert primitives in `MergeHiveSchemaWithAvro` but it can be used as an independent visitor if required

cc: @wmoustafa @rzhang10


hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveSchemaWithPartnerVisitor.java
```diff
@@ -0,0 +1,127 @@
...
+    FP fieldPartner(P partnerStruct, String fieldName);
+
+    P fieldType(FP partnerField);
+
+    P mapKeyPartner(P partnerMap);
+
+    P mapValuePartner(P partnerMap);
+
+    P listElementPartner(P partnerList);
+  }
```
wmoustafa, line 63: I would remove the name `partner` from the method names and variable names (and in the implementation too) since it is implied from the class name. We might also bring those methods directly to `HiveSchemaWithPartnerVisitor`. We can do this after check-in.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/MergeHiveSchemaWithAvro.java
```diff
@@ -0,0 +1,197 @@
...
+    // TODO: Check if schema is compatible with partner
+    //       Also do type promotion if required, schema = string & partner = enum, schema = bytes & partner = fixed, etc
+    return schema;
+  }
+
+  /**
+   * A {@link PartnerAccessors} which matches the requested field from a partner Avro struct by case insensitive
+   * field name match
+   */
+  private static class AvroPartnerAccessor implements PartnerAccessors<Schema, Schema.Field> {
```
wmoustafa, line 144: Nit: `AvroPartnerAccessor implements PartnerAccessor` (without `s` at the end) reads better in my opinion. We can do that after check-in as well.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.hadoop.hive.serde2.typeinfo.ListTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.MapTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.PrimitiveTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.StructTypeInfo;
+import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
+import org.apache.iceberg.avro.AvroSchemaUtil;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.codehaus.jackson.node.JsonNodeFactory;
+
+class HiveTypeToAvroType extends HiveTypeUtil.HiveSchemaVisitor<Schema> {
```
wmoustafa, line 36: If this is not used anywhere, the code will be much simpler if we just leverage a method for the primitive conversion (as opposed to full-fledged class with complex type support).

wmoustafa (approved):
Thanks for this patch. Looks good overall.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/MergeHiveSchemaWithAvro.java
shardulm94: Makes sense

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveTypeToAvroType.java
shardulm94: Done

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveSchemaWithPartnerVisitor.java
shardulm94: Lets followup on this after the check-in. I may not have the same opinion and don't want to delay the rollout since this seems like a minor issue.

wmoustafa (approved):

PR #59, opened 2021-03-16
Fix empty struct type in selectNot call

rzhang10:
Tested the table with the empty struct schema using spark-shell, the table could now be read without exception.


shardulm94:
@rzhang10 Can you file this upstream? Since this touches the API module, it will be helpful to not diverge from upstream.

PR #63, opened 2021-04-09
 [#2039] Support default value semantics - first PR: API changes

shenodaguirguis:
Adding defaultValue and its APIs to Iceberg Types.NestedField to support default values semantics ([issue #2039](https://github.com/apache/iceberg/issues/2039)). 
This is PR 1 of 5. 
For detailed problem and solution check [defaultValueSupport.md](https://github.com/shenodaguirguis/iceberg/blob/master/defaultValueSupport.md)

### DIFF
--- a/api/src/main/java/org/apache/iceberg/types/Types.java
+++ b/api/src/main/java/org/apache/iceberg/types/Types.java
@@ -20,6 +20,7 @@
 package org.apache.iceberg.types;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
@@ -414,42 +415,108 @@ public class Types {
 
   public static class NestedField implements Serializable {
     public static NestedField optional(int id, String name, Type type) {
-      return new NestedField(true, id, name, type, null);
+      return new NestedField(true, id, name, type, null, null);
     }
 
     public static NestedField optional(int id, String name, Type type, String doc) {
-      return new NestedField(true, id, name, type, doc);
+      return new NestedField(true, id, name, type, null, doc);
+    }
+
+    public static NestedField optional(int id, String name, Type type, Object defaultValue, String doc) {
+      return new NestedField(true, id, name, type, defaultValue, doc);
     }
 
     public static NestedField required(int id, String name, Type type) {
-      return new NestedField(false, id, name, type, null);
+      return new NestedField(false, id, name, type, null, null);
     }
 
     public static NestedField required(int id, String name, Type type, String doc) {
-      return new NestedField(false, id, name, type, doc);
+      return new NestedField(false, id, name, type, null, doc);
+    }
+
+    public static NestedField required(int id, String name, Type type, Object defaultValue, String doc) {
+      validateDefaultValueForRequiredField(defaultValue, name);
+      return new NestedField(false, id, name, type, defaultValue, doc);
     }
 
     public static NestedField of(int id, boolean isOptional, String name, Type type) {
-      return new NestedField(isOptional, id, name, type, null);
+      return new NestedField(isOptional, id, name, type, null, null);
     }
 
     public static NestedField of(int id, boolean isOptional, String name, Type type, String doc) {
-      return new NestedField(isOptional, id, name, type, doc);
+      return new NestedField(isOptional, id, name, type, null, doc);
+    }
+
+    public static NestedField of(int id, boolean isOptional, String name, Type type, Object defaultValue, String doc) {
+      return new NestedField(isOptional, id, name, type, defaultValue, doc);
+    }
+
+    private static void validateDefaultValueForRequiredField(Object defaultValue, String fieldName) {
+      Preconditions.checkArgument(defaultValue != null,
+          "Cannot create NestedField with a null default for the required field: " +  fieldName);
+    }
+
+    private static void validateDefaultValue(Object defaultValue, Type type) {
+      if (defaultValue == null) {
+        return;
+      }
+      switch (type.typeId()) {
+        case STRUCT:
+          Preconditions.checkArgument(List.class.isInstance(defaultValue),
+              "defaultValue should be a List of Objects for StructType");
+          if (defaultValue == null) {
+            return;
+          }
+          List<Object> defaultList = (List) defaultValue;
+          Preconditions.checkArgument(defaultList.size() == type.asStructType().fields().size());
+          for (int i = 0; i < defaultList.size(); i++) {
+            NestedField.validateDefaultValue(defaultList.get(i), type.asStructType().fields().get(i).type);
+          }
+          break;
+        case LIST:
+          Preconditions.checkArgument(defaultValue instanceof ArrayList,
+              "defaultValue should be an ArrayList of Objects, for ListType");
+          List<Object> defaultArrayList = (ArrayList<Object>) defaultValue;
+          if (defaultArrayList == null || defaultArrayList.size() == 0) {
+            return;
+          }
+          defaultArrayList.forEach(dv -> NestedField.validateDefaultValue(dv, type.asListType().elementField.type));
+          break;
+        case MAP:
+          Preconditions.checkArgument(Map.class.isInstance(defaultValue),
+              "defaultValue should be an instance of Map for MapType");
+          Map<Object, Object> defaultMap = (Map<Object, Object>) defaultValue;
+          if (defaultMap == null || defaultMap.isEmpty()) {
+            return;
+          }
+          for (Map.Entry e : defaultMap.entrySet()) {
+            NestedField.validateDefaultValue(e.getKey(), type.asMapType().keyField.type);
+            NestedField.validateDefaultValue(e.getValue(), type.asMapType().valueField.type);
+          }
+          break;
+        default:
+          Preconditions.checkArgument(type.typeId().javaClass().isInstance(defaultValue),
+              "defaultValue should be of same java class of the type, defaultValue class: " + defaultValue.getClass() +
+              ", type class: " + type.typeId().javaClass());
+      }
     }
 
     private final boolean isOptional;
     private final int id;
     private final String name;
     private final Type type;
+    private final Object defaultValue;
     private final String doc;
 
-    private NestedField(boolean isOptional, int id, String name, Type type, String doc) {
+    private NestedField(boolean isOptional, int id, String name, Type type, Object defaultValue, String doc) {
       Preconditions.checkNotNull(name, "Name cannot be null");
       Preconditions.checkNotNull(type, "Type cannot be null");
+      validateDefaultValue(defaultValue, type);
       this.isOptional = isOptional;
       this.id = id;
       this.name = name;
       this.type = type;
+      this.defaultValue = defaultValue;
       this.doc = doc;
     }
 
@@ -461,7 +528,7 @@ public class Types {
       if (isOptional) {
         return this;
       }
-      return new NestedField(true, id, name, type, doc);
+      return new NestedField(true, id, name, type, defaultValue, doc);
     }
 
     public boolean isRequired() {
@@ -472,7 +539,15 @@ public class Types {
       if (!isOptional) {
         return this;
       }
-      return new NestedField(false, id, name, type, doc);
+      return new NestedField(false, id, name, type, defaultValue, doc);
+    }
+
+    public boolean hasDefaultValue() {
+      return defaultValue != null;
+    }
+
+    public Object getDefaultValue() {
+      return defaultValue;
     }
 
     public int fieldId() {
@@ -495,6 +570,7 @@ public class Types {
     public String toString() {
       return String.format("%d: %s: %s %s",
           id, name, isOptional ? "optional" : "required", type) +
+          (hasDefaultValue() ? ", default value: " + defaultValue + ", " : "") +
           (doc != null ? " (" + doc + ")" : "");
     }
 
@@ -513,6 +589,8 @@ public class Types {
         return false;
       } else if (!name.equals(that.name)) {
         return false;
+      } else if (!Objects.equals(defaultValue, that.defaultValue)) {
+        return false;
       } else if (!Objects.equals(doc, that.doc)) {
         return false;
       }
@@ -521,7 +599,8 @@ public class Types {
 
     @Override
     public int hashCode() {
-      return Objects.hash(NestedField.class, id, isOptional, name, type);
+      return hasDefaultValue() ? Objects.hash(NestedField.class, id, isOptional, name, type, defaultValue) :
+          Objects.hash(NestedField.class, id, isOptional, name, type);
     }
   }
 
@@ -739,7 +818,6 @@ public class Types {
       } else if (!(o instanceof ListType)) {
         return false;
       }
-
       ListType listType = (ListType) o;
       return elementField.equals(listType.elementField);
     }
--- a/api/src/test/java/org/apache/iceberg/types/TestDefaultValuesForContainerTypes.java
+++ b/api/src/test/java/org/apache/iceberg/types/TestDefaultValuesForContainerTypes.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -472,7 +494,15 @@ public NestedField asRequired() {
...
-      return new NestedField(false, id, name, type, doc);
+      return new NestedField(false, id, name, type, defaultValue, doc);
+    }
+
+    public boolean hasDefaultValue() {
+      return defaultValue != null;
+    }
+
+    public Object getDefaultValue() {
+      return defaultValue;
```
wmoustafa, line 505: I think dealing with null-default as no-default will work, but from an Avro schema point of view, there is a distinction. Do we want to maintain that distinction, or say that default values in Iceberg are non-null default values only? I am okay with the latter as long as we do not end up doing complex checks to disambiguate between null-default and no-default. This will happen if we do not need to distinguish anywhere to begin with.

wmoustafa: Also, I am okay with proceeding with this for now and re-evaluating later, but it is something to think about.

shenodaguirguis: Thanks for the review @wmoustafa! 
null-default values today are mapped to optional NestedField (with no default). This shall remain the same, given that for optional NestedFields with no defaults, if the field is not manifested then null will be used. In other words, the mapping will be as follows (which shall become clearer in next PR, which handles the schema mapping/conversion):
```
avro field                 ->  iceberg nested field    | what happens when field is not manifested 
------------------------------------------------------------------------------------------------
optional, no default       ->  optional, no default.   | null is used
required, no default       ->  required, no default.   | error
optional, non-null default ->  optional, with default  | default value is used
required, non-null default ->  required, with default  | default value is used
optional, null default     ->  optional, no default    | null is used
required, null default     ->  optional, no default    | null (same as avro's default value) value is used
```

wmoustafa: I had the same table in my scratch notes to reason about this :) Yes, agreed it should work as I mentioned, just discussing this from a semantic point of view and code complexity down the line, but seems the latter will not be an issue, and the former should not be an issue as long as the latter is not.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -414,42 +414,71 @@ public int hashCode() {
...
     public static NestedField of(int id, boolean isOptional, String name, Type type, String doc) {
-      return new NestedField(isOptional, id, name, type, doc);
+      return new NestedField(isOptional, id, name, type, null, doc);
     }
 
+    public static NestedField of(int id, boolean isOptional, String name, Type type, Object defaultValue, String doc) {
+      return new NestedField(isOptional, id, name, type, defaultValue, doc);
+    }
+
+    private static void validateDefaultValueForRequiredField(Object defaultValue) {
```
shardulm94, line 462: Might be better to just move this check into the constructor of `NestedField`.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -414,42 +414,71 @@ public int hashCode() {
...
+      Preconditions.checkNotNull(defaultValue, "defaultValue cannot be null for a required field");
+    }
+
+
     private final boolean isOptional;
     private final int id;
     private final String name;
     private final Type type;
+    private final Object defaultValue;
```
shardulm94, line 471: What concrete Java types are supported for this `Object`. Can we add some validation to ensure the default value objects are compatible with their data types?

shardulm94, line 471: Are default values supported for container types like struct/map/list?

api/src/main/java/org/apache/iceberg/types/Types.java
shenodaguirguis: makes sense

api/src/main/java/org/apache/iceberg/types/Types.java
shenodaguirguis: Object Java type should matches that of the Type, added validation to the constructor

api/src/main/java/org/apache/iceberg/types/Types.java
shenodaguirguis: so, we are using defaultValue = null to represent no-defaultValue. 
The validation was meant to avoid the case when the user calls an API that passes defaultValue and sets the defaultValue to null in error (to avoid behavior that might be unexpected by the user). However, we can say it is legal to set defaultValue to null, which is equivalent to calling and API with no defaultValue, and we leave it to the user's responsibility to user the right APIs and parameters values (with adding java docs regarding defaultValue to the corresponding APIs). Otherwise, the validation will have to take place in the relevant APIs, not the constructor (because at constructor level it is legal) OR, we will have to use a different mechanism to represent no default value. For example, adding a _hasDefault flag, which is not ideal (extra effort to API callers) or use Optional which is not recommended for parameters.

api/src/test/java/org/apache/iceberg/types/TestNestedFiledDefaultValues.java
```diff
@@ -0,0 +1,71 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg.types;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestNestedFiledDefaultValues {
```
shardulm94, line 26: Typo: Filed -> Field

api/src/main/java/org/apache/iceberg/types/Types.java
shenodaguirguis: Avro 1.8 supports default values for record, map, array, and enum
from https://avro.apache.org/docs/1.8.1/spec.html#schema_record
```
avro type | json type | example
-- | -- | --
...
record | object | {"a": 1}
enum | string | "FOO"
array | array | [1]
map | object | {"a": 1}
```

wmoustafa (approved):

shardulm94:
I feel we should wait for some time to get early feedback from the community to make sure we agree on a high level. This should help reduce churn in our codebase. Most of my comments in https://github.com/apache/iceberg/pull/2496#pullrequestreview-639438937 are valid for the latest set of changes pushed by @shenodaguirguis and I hope they can be addressed before we merge this internally.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -414,42 +415,96 @@ public int hashCode() {
...
+            "defaultValue should be a List of Objects for StructType");
+        StructType.validateDefaultValue((List) defaultValue, type.asStructType().fieldList);
+      } else if (type.isListType()) {
+        Preconditions.checkArgument(defaultValue.getClass().isArray(),
+            "defaultValue should be an array of Objects for ListType");
+        ListType.validateDefaultValue((Object[]) defaultValue, type.asListType().elementField);
+      } else if (type.isMapType()) {
+        Preconditions.checkArgument(Map.class.isInstance(defaultValue),
+            "defaultValue should be an instance of Map for MapType");
+        MapType.validateDefaultValue((Map<Object, Object>) defaultValue,
```
wmoustafa, line 483: Can we eliminate `XType.validateDefaultValue` and have this method call itself recursively?

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -759,26 +885,41 @@ public int hashCode() {
 
   public static class MapType extends NestedType {
     public static MapType ofOptional(int keyId, int valueId, Type keyType, Type valueType) {
+      return ofOptional(keyId, valueId, keyType, valueType, null);
+    }
+
+    public static MapType ofOptional(int keyId, int valueId, Type keyType, Type valueType,
+        Map<Object, Object> defaultValue) {
```
wmoustafa, line 892: I am confused why we need to pass a default value here. Are not the parameters of this method already at the level of the Map keys and values? Passing a Map-level concept seems incompatible with what this method is trying to achieve.

wmoustafa: Also, this method (while new) is not called anywhere in this patch.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -858,16 +1007,32 @@ public boolean equals(Object o) {
...
+      if (!keyField.equals(that.keyField)) {
+        return false;
+      }
+      return valueField.equals(that.valueField);
     }
 
     @Override
     public int hashCode() {
-      return Objects.hash(MapType.class, keyField, valueField);
+      return hasDefaultValue() ? Objects.hash(MapType.class, keyField, valueField, defaultValue) :
```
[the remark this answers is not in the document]
wmoustafa: Answer to this question depends on https://github.com/linkedin/iceberg/pull/63#discussion_r623705887.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -608,12 +691,26 @@ public boolean equals(Object o) {
...
     @Override
     public int hashCode() {
-      return Objects.hash(NestedField.class, Arrays.hashCode(fields));
+      return hasDefaultValue() ? Objects.hash(NestedField.class, Arrays.hashCode(fields), defaultValue.hashCode()) :
+          Objects.hash(NestedField.class, Arrays.hashCode(fields));
+    }
+
+    private static void validateDefaultValue(List<Object> defaultValue, List<NestedField> fields) {
+      if (defaultValue == null) {
+        return;
```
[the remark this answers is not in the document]
wmoustafa: Answer to this question depends on https://github.com/linkedin/iceberg/pull/63#discussion_r623705887.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -608,12 +691,26 @@ public boolean equals(Object o) {
...
+        return false;
+      }
       return Arrays.equals(fields, that.fields);
     }
 
     @Override
     public int hashCode() {
-      return Objects.hash(NestedField.class, Arrays.hashCode(fields));
+      return hasDefaultValue() ? Objects.hash(NestedField.class, Arrays.hashCode(fields), defaultValue.hashCode()) :
+          Objects.hash(NestedField.class, Arrays.hashCode(fields));
```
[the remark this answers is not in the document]
wmoustafa: Answer to this question depends on https://github.com/linkedin/iceberg/pull/63#discussion_r623705887.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -830,6 +971,13 @@ public boolean isValueOptional() {
       return valueField.isOptional;
     }
 
+    public Map<Object, Object> getDefaultValue() {
+      return Collections.unmodifiableMap(defaultValue);
+    }
+
+    public boolean hasDefaultValue() {
+      return defaultValue != null;
+    }
```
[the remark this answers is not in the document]
wmoustafa: Answer to this question depends on https://github.com/linkedin/iceberg/pull/63#discussion_r623705887.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -858,16 +1007,32 @@ public boolean equals(Object o) {
         return false;
       }
 
-      MapType mapType = (MapType) o;
-      if (!keyField.equals(mapType.keyField)) {
+      MapType that = (MapType) o;
+      if (hasDefaultValue() && !defaultValue.equals(that.defaultValue)) {
```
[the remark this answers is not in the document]
wmoustafa: Answer to this question depends on https://github.com/linkedin/iceberg/pull/63#discussion_r623705887.

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -414,42 +415,96 @@ public int hashCode() {
...
-      return new NestedField(true, id, name, type, doc);
+      return new NestedField(true, id, name, type, null, doc);
+    }
+
+    public static NestedField optional(int id, String name, Type type, Object defaultValue, String doc) {
+      return new NestedField(true, id, name, type, defaultValue, doc);
+    }
+
+    public static NestedField optionalWithDefault(int id, String name, Type type, Object defaultValue) {
+      return new NestedField(true, id, name, type, defaultValue, null);
```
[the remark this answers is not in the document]
shenodaguirguis: compiler cannot distinguish Object (defaultVlaue) from String (doc) if both APIs have the same name. However, based on our offline discussion, I verified that it is sufficient to maintain the default value in the NestedField level. So, StructType, ListType and MapType changes are revered

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -414,42 +415,96 @@ public int hashCode() {
...
+      return new NestedField(isOptional, id, name, type, null, doc);
+    }
+
+    public static NestedField of(int id, boolean isOptional, String name, Type type, Object defaultValue, String doc) {
+      return new NestedField(isOptional, id, name, type, defaultValue, doc);
+    }
+
+    private static void validateDefaultValueForRequiredField(Object defaultValue, String fieldName) {
+      Preconditions.checkArgument(defaultValue != null,
+          "Cannot create NestedField with a null default for a required field " +  fieldName);
```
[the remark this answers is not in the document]
shenodaguirguis: here, it is a required field and we want to make sure the defaultValue  is not null, because (based on the table [here](https://github.com/linkedin/iceberg/pull/63#discussion_r612030559) this not allowed. In the constructor itself, however, null defaultValue is valid and it means no-defaultValue

api/src/main/java/org/apache/iceberg/types/Types.java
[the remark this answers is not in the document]
shenodaguirguis: I am using `hasDefaultValue()` whenever possible (in non-static methods)

api/src/main/java/org/apache/iceberg/types/Types.java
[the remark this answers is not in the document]
shenodaguirguis: yes, done

api/src/main/java/org/apache/iceberg/types/Types.java
shenodaguirguis: that is true. Removed defaultValue from Map/List/Struct level, it is sufficient to maintain it at the NestedField level.

api/src/main/java/org/apache/iceberg/types/Types.java
shenodaguirguis: done

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -414,42 +415,96 @@ public int hashCode() {
...
+    }
+
+    private static void validateDefaultValue(Object defaultValue, Type type) {
+      if (defaultValue == null) {
+        return;
+      }
+      if (type.isStructType()) {
```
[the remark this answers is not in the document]
shenodaguirguis: done

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -739,9 +827,8 @@ public boolean equals(Object o) {
       } else if (!(o instanceof ListType)) {
         return false;
       }
-
-      ListType listType = (ListType) o;
-      return elementField.equals(listType.elementField);
+      ListType that = (ListType) o;
+      return elementField.equals(that.elementField);
```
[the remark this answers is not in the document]
shenodaguirguis: reverted

api/src/main/java/org/apache/iceberg/types/Types.java
```diff
@@ -858,11 +945,11 @@ public boolean equals(Object o) {
...
       }
 
-      MapType mapType = (MapType) o;
-      if (!keyField.equals(mapType.keyField)) {
+      MapType that = (MapType) o;
+      if (!keyField.equals(that.keyField)) {
         return false;
       }
-      return valueField.equals(mapType.valueField);
+      return valueField.equals(that.valueField);
```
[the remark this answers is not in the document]
shenodaguirguis: reverted

wmoustafa (approved):
LGTM

PR #64, opened 2021-04-12
Stop using serdeToFileFormat to unblock formats other than Avro or Orc

rzhang10:
At LinkedIn, we have internally custom format that is non Avro nor Orc, but the previous code used function `serdeToFileFormat` which when called will block non Avro or Orc format being read. This is a simple fix to unblock.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
@@ -61,7 +61,8 @@ class LegacyHiveTableUtils {
     if (avroSchema != null) {
       String serde = table.getSd().getSerdeInfo().getSerializationLib();
       org.apache.avro.Schema finalAvroSchema;
-      if (serde.equals("org.apache.hadoop.hive.serde2.avro.AvroSerDe") || HasDuplicateLowercaseColumnNames.visit(avroSchema)) {
+      if (serde.equals("org.apache.hadoop.hive.serde2.avro.AvroSerDe") ||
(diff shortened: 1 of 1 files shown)


wmoustafa (approved):

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -59,9 +59,9 @@ static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
     org.apache.avro.Schema avroSchema = schemaStr != null ? new org.apache.avro.Schema.Parser().parse(schemaStr) : null;
     Schema schema;
     if (avroSchema != null) {
-      FileFormat format = serdeToFileFormat(table.getSd().getSerdeInfo().getSerializationLib());
+      String serde = table.getSd().getSerdeInfo().getSerializationLib();
       org.apache.avro.Schema finalAvroSchema;
-      if (format.equals(FileFormat.AVRO) || HasDuplicateLowercaseColumnNames.visit(avroSchema)) {
+      if (serde.equals("org.apache.hadoop.hive.serde2.avro.AvroSerDe") || HasDuplicateLowercaseColumnNames.visit(avroSchema)) {
```
funcheetah, line 64: Could this be defined as constant? I notice "org.apache.hadoop.hive.serde2.avro.AvroSerDe" also shows up in other places.

rzhang10: I think it's fine to use it as is, since there are not other code re-use it.

PR #65, opened 2021-04-27
Support reading Avro complex union types

funcheetah:
This PR supports reading Avro complex union types as Iceberg Struct type via `SparkAvroReader`. 
For example, 
`[ "int", "string" ] => required struct (
0 tag_0: optional int,
1 tag_1: optional string)
`
`
[ "null", "int", "string" ] => optional struct (
0 tag_0: optional int,
1 tag_1: optional string)
`

Testing done:
1. unit test
2. integration test via spark reading table with non-option union

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
@@ -130,7 +130,16 @@ public class AvroSchemaUtil {
     return false;
   }
 
-  public static boolean isNonOptionalUnionOptional(Schema schema) {
+  /**
+   * This method decides whether a schema is of type union and is complex union and is optional
+   *
+   * Complex union: the number of options in union not equals to 2
+   * Optional: null is present in union
+   *
+   * @param schema
+   * @return true if schema is complex union and it is optional
+   */
+  public static boolean isOptionalComplexUnion(Schema schema) {
     if (schema.getType() == UNION && schema.getTypes().size() != 2) {
       for (Schema type : schema.getTypes()) {
         if (type.getType() == Schema.Type.NULL) {
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
@@ -130,6 +130,18 @@ public class AvroSchemaUtil {
     return false;
   }
 
+  public static boolean isNonOptionalUnionOptional(Schema schema) {
+    if (schema.getType() == UNION && schema.getTypes().size() != 2) {
+      for (Schema type : schema.getTypes()) {
+        if (type.getType() == Schema.Type.NULL) {
+          return true;
+        }
+      }
+    }
+
+    return false;
+  }
+
   public static Schema toOption(Schema schema) {
     if (schema.getType() == UNION) {
       Preconditions.checkArgument(isOptionSchema(schema),
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
@@ -136,7 +136,7 @@ public class AvroSchemaUtil {
    * Complex union: the number of options in union not equals to 2
    * Optional: null is present in union
    *
-   * @param schema
+   * @param schema input schema
    * @return true if schema is complex union and it is optional
    */
   public static boolean isOptionalComplexUnion(Schema schema) {
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -49,7 +49,8 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
             visit(map != null ? map.valueType() : null, schema.getValueType(), visitor));
 
       default:
-        return visitor.primitive(iType != null ? iType.asPrimitiveType() : null, schema);
+        return visitor.primitive(null, schema);
+        // return visitor.primitive(iType != null ? iType.asPrimitiveType() : null, schema);
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -85,10 +85,10 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
       if (branch.getType() == Schema.Type.NULL) {
         options.add(visit((Type) null, branch, visitor));
       } else {
-        if (type.isStructType()) {
-          options.add(visit(type.asStructType().fields().get(index).type(), branch, visitor));
-        } else {
+        if (AvroSchemaUtil.isOptionSchema(union)) {
           options.add(visit(type, branch, visitor));
+        } else {
+          options.add(visit(type.asStructType().fields().get(index).type(), branch, visitor));
         }
         index++;
       }
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -49,10 +49,7 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
             visit(map != null ? map.valueType() : null, schema.getValueType(), visitor));
 
       default:
-        if (iType == null || iType.isStructType()) {
-          return visitor.primitive(null, schema);
-        }
-        return visitor.primitive(iType.asPrimitiveType(), schema);
+        return visitor.primitive(iType != null ? iType.asPrimitiveType() : null, schema);
     }
   }
 
@@ -82,11 +79,18 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
   private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisitor<T> visitor) {
     List<Schema> types = union.getTypes();
     List<T> options = Lists.newArrayListWithExpectedSize(types.size());
+
+    int index = 0;
     for (Schema branch : types) {
       if (branch.getType() == Schema.Type.NULL) {
         options.add(visit((Type) null, branch, visitor));
       } else {
-        options.add(visit(type, branch, visitor));
+        if (type.isStructType()) {
+          options.add(visit(type.asStructType().fields().get(index).type(), branch, visitor));
+        } else {
+          options.add(visit(type, branch, visitor));
+        }
+        index++;
       }
     }
     return visitor.union(type, union, options);
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -49,8 +49,10 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
             visit(map != null ? map.valueType() : null, schema.getValueType(), visitor));
 
       default:
-        return visitor.primitive(null, schema);
-        // return visitor.primitive(iType != null ? iType.asPrimitiveType() : null, schema);
+        if (iType == null || iType.isStructType()) {
+          return visitor.primitive(null, schema);
+        }
+        return visitor.primitive(iType.asPrimitiveType(), schema);
     }
   }
 
--- a/core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
+++ b/core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
@@ -96,7 +96,7 @@ class SchemaToType extends AvroSchemaVisitor<Type> {
       Object defaultValue = field.hasDefaultValue() && !(field.defaultVal() instanceof JsonProperties.Null) ?
           field.defaultVal() : null;
 
-      if (AvroSchemaUtil.isOptionSchema(field.schema()) || AvroSchemaUtil.isNonOptionalUnionOptional(field.schema())) {
+      if (AvroSchemaUtil.isOptionSchema(field.schema()) || AvroSchemaUtil.isOptionalComplexUnion(field.schema())) {
         newFields.add(Types.NestedField.optional(fieldId, field.name(), fieldType, defaultValue, null));
       } else if (defaultValue != null) {
         newFields.add(Types.NestedField.required(fieldId, field.name(), fieldType, defaultValue, null));
@@ -111,7 +111,7 @@ class SchemaToType extends AvroSchemaVisitor<Type> {
   @Override
   public Type union(Schema union, List<Type> options) {
     if (AvroSchemaUtil.isOptionSchema(union)) {
-      // optional union
+      // Optional simple union
       // records, arrays, and maps will check nullability later
       if (options.get(0) == null) {
         return options.get(1);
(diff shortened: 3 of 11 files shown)


core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
```diff
@@ -104,13 +104,26 @@ public Type record(Schema record, List<String> names, List<Type> fieldTypes) {
...
+      List<Types.NestedField> newFields = Lists.newArrayListWithExpectedSize(options.size());
+
+      int tagIndex = 1;
+      for (Type type : options) {
+        if (type != null) {
+          newFields.add(Types.NestedField.optional(allocateId(), "tag_" + tagIndex++, type));
+        }
+      }
+
+      return Types.StructType.of(newFields);
```
shenodaguirguis, line 126: not sure I follow the reasoning why struct is the appropriate way to represent non-optional union.  Should we introduce a new Types.unionType (that extends struct may be)? Because using struct per se (with optional fields) is somewhat confusing

core/src/test/java/org/apache/iceberg/avro/TestAvroNonOptionalUnion.java
```diff
@@ -0,0 +1,91 @@
...
+
+package org.apache.iceberg.avro;
+
+import org.apache.avro.Schema;
+import org.apache.avro.SchemaBuilder;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestAvroNonOptionalUnion {
```
shenodaguirguis, line 28: can we add tests for invalid cases as well?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -285,4 +291,36 @@ protected void set(InternalRow struct, int pos, Object value) {
...
+    }
+
+    @Override
+    public InternalRow read(Decoder decoder, Object reuse) throws IOException {
+      InternalRow struct = new GenericInternalRow(readers.length);
+      int index = decoder.readIndex();
+      Object value = this.readers[index].read(decoder, reuse);
+
+      struct.update(index, value);
+      for (int i = 0; i < readers.length && i != index; i += 1) {
```
shenodaguirguis, line 319: should we do this for all i before we set the value at index? This loop is setting null all fields before the value, but not after...

rzhang10 (commented):
Just want to add a general comment here, so I see this is to support reading unions in avro format data in spark, are we also going to support reading union in orc format in spark? If that's the case, I feel the scope will expand since we are merging hive and avro schema literal as the iceberg schema for orc format, and if the hive schema has a union it will throw and exception as unsupported.

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
```diff
@@ -104,13 +104,26 @@ public Type record(Schema record, List<String> names, List<Type> fieldTypes) {
...
+        return options.get(1);
+      } else {
+        return options.get(0);
+      }
     } else {
-      return options.get(0);
+      // non-option union
+      List<Types.NestedField> newFields = Lists.newArrayListWithExpectedSize(options.size());
+
+      int tagIndex = 1;
```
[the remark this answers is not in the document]
funcheetah: Yes, it should start at `0` for compatibility

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
funcheetah: Discussed offline with @shenodaguirguis . The reasoning is documented in design doc. The non-optional union can be represented with struct type as the following example
`
[ “null”, “int”, “string” ] => optional struct (
1 tag_0: optional int,
2 tag_1: optional string)
`

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
funcheetah: Thanks for catching this. Change to explicitly set all fields to null before setting value at index.

core/src/test/java/org/apache/iceberg/avro/TestAvroNonOptionalUnion.java
funcheetah: Add more tests including non-option union.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -49,7 +49,10 @@
             visit(map != null ? map.valueType() : null, schema.getValueType(), visitor));
 
       default:
-        return visitor.primitive(iType != null ? iType.asPrimitiveType() : null, schema);
+        if (iType == null || iType.isStructType()) {
```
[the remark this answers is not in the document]
funcheetah: When `visitUnion`, each element of UNION will be visited along with iType as StructType. Thus, we need to handle iType properly. Discussed offline, we decide to move the StructType handling logic to `visitUnion` so that no change need to be made here.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -49,7 +49,10 @@
...
+          return visitor.primitive(null, schema);
```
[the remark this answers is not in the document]
funcheetah: When visitUnion, each element of UNION will be visited along with iType as StructType. Thus, we need to handle iType properly. Discussed offline, we decide to move the StructType handling logic to visitUnion so that no change need to be made here.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
```diff
@@ -130,6 +130,18 @@ public static boolean isOptionSchema(Schema schema) {
     return false;
   }
 
+  public static boolean isNonOptionalUnionOptional(Schema schema) {
+    if (schema.getType() == UNION && schema.getTypes().size() != 2) {
```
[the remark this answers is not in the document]
shenodaguirguis: good catch @rzhang10 ! While it is meaningless to have a single-type in a union, the avro `Schema.UnionSchema` constructor does not perform any checks on the number of types, so we definitely need to guard against such cases (throw if size < 2 ?).

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -79,11 +79,18 @@
...
       if (branch.getType() == Schema.Type.NULL) {
         options.add(visit((Type) null, branch, visitor));
       } else {
-        options.add(visit(type, branch, visitor));
+        if (AvroSchemaUtil.isOptionSchema(union)) {
+          options.add(visit(type, branch, visitor));
+        } else {
+          options.add(visit(type.asStructType().fields().get(index).type(), branch, visitor));
+        }
+        index++;
```
shenodaguirguis, line 93: nit: we now maintain index outside the loop, and check if union is option (line 88) every time in the loop.  May be  use one loop for both branches and fields, and re-org to hoist up the union check (if that is feasible)?

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
```diff
@@ -104,13 +104,26 @@ public Type record(Schema record, List<String> names, List<Type> fieldTypes) {
...
-      return options.get(1);
+    if (AvroSchemaUtil.isOptionSchema(union)) {
+      // optional union
+      // records, arrays, and maps will check nullability later
+      if (options.get(0) == null) {
+        return options.get(1);
+      } else {
+        return options.get(0);
+      }
     } else {
```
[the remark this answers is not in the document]
shenodaguirguis: +1

shenodaguirguis (commented):
Thanks for the changes @funcheetah ! 
LGTM in general. Other than the nits, the only - small - open issue I see is to guard against the wrong corner case of union type with < 2 types which @rzhang10  pointed out

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
```diff
@@ -130,6 +130,18 @@ public static boolean isOptionSchema(Schema schema) {
     return false;
   }
 
+  public static boolean isNonOptionalUnionOptional(Schema schema) {
```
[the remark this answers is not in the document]
funcheetah: Option union only refers to `[null, type]`. While `[int, string]`, `[int]`, `[null, int, string]` etc are all non-option unions. What this method checks is whether a non-option union is optional. I think the original name is more accurate.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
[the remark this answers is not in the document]
funcheetah: Yeah, it can happen. For example, `[null]`  and `[int]` will pass the check. However, those are valid Avro unions and Avro does not prevent users from creating those schemas. In addition, our reader will works with those single unions. Maybe it is better to support those than guard it? We can let the user to decide whether it is useful or not.

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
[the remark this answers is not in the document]
funcheetah: Yes, it will create an empty struct. However, it is a valid Avro schema and our reader can read it without issue. Should we support it instead of guard against it?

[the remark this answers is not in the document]
shenodaguirguis: thanks for validating this Wenye. I am fine with not guarding against it, given that avro schema also does not guard against this corner case.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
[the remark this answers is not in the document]
shenodaguirguis: I see. `Option` is an Iceberg (bad) term  == Optional with 2 types only.
Would reversing logic and renaming to `isNonOptionUniontRequired` make more sense? I will leave it to your preference. But if you keep it as `isNonOptionUnionOptional`, please clarify in a java-doc the difference between optional and option/non-option

shenodaguirguis (approved):

core/src/test/java/org/apache/iceberg/avro/TestAvroNonOptionalUnion.java
```diff
@@ -0,0 +1,129 @@
...
```
wmoustafa, line 28: `TestAvroComplexUnion` Please change everywhere else.

core/src/test/java/org/apache/iceberg/avro/TestAvroNonOptionalUnion.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Test
+  public void testNonOptionUnionNonNullable() {
```
wmoustafa, line 31: `testRequiredComplexUnion`

core/src/test/java/org/apache/iceberg/avro/TestAvroNonOptionalUnion.java
```diff
@@ -0,0 +1,129 @@
...
+
+    org.apache.iceberg.Schema icebergSchema = AvroSchemaUtil.toIceberg(avroSchema);
+    String expectedIcebergSchema = "table {\n" +
+        "  0: unionCol: required struct<1: tag_0: optional int, 2: tag_1: optional string>\n" + "}";
+
+    Assert.assertEquals(expectedIcebergSchema, icebergSchema.toString());
+  }
+
+  @Test
+  public void testNonOptionUnionNullable() {
```
wmoustafa, line 52: `testOptionalComplexUnion`

core/src/test/java/org/apache/iceberg/avro/TestAvroNonOptionalUnion.java
```diff
@@ -0,0 +1,129 @@
...
+
+    org.apache.iceberg.Schema icebergSchema = AvroSchemaUtil.toIceberg(avroSchema);
+    String expectedIcebergSchema =
+        "table {\n" + "  0: unionCol: optional struct<1: tag_0: optional int, 2: tag_1: optional string>\n" + "}";
+
+    Assert.assertEquals(expectedIcebergSchema, icebergSchema.toString());
+  }
+
+  @Test
+  public void testSingleOptionUnion() {
```
wmoustafa, line 75: `testSingleComponentUnion`

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -0,0 +1,163 @@
...
+import org.junit.rules.TemporaryFolder;
+
+
+public class TestSparkAvroUnions {
+
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  @Test
+  public void writeAndValidateNonOptionUnionNonNullable() throws IOException {
```
wmoustafa, line 48: `RequiredComplexUnion`

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -0,0 +1,163 @@
...
+    try (AvroIterable<InternalRow> reader = Avro.read(Files.localInput(testFile))
+        .createReaderFunc(SparkAvroReader::new)
+        .project(expectedSchema)
+        .build()) {
+      rows = Lists.newArrayList(reader);
+    }
+  }
+
+  @Test
+  public void writeAndValidateNonOptionUnionNullable() throws IOException {
```
wmoustafa, line 87: `OptionalComplexUnion`

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -0,0 +1,163 @@
...
+    try (AvroIterable<InternalRow> reader = Avro.read(Files.localInput(testFile))
+        .createReaderFunc(SparkAvroReader::new)
+        .project(expectedSchema)
+        .build()) {
+      rows = Lists.newArrayList(reader);
+    }
+  }
+
+  @Test
+  public void writeAndValidateSingleOptionUnion() throws IOException {
```
wmoustafa, line 128: `SingleComponent`

funcheetah: Done.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
funcheetah: Done.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
funcheetah: Done.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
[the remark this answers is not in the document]
funcheetah: Thanks for the suggestion. Renamed it to `isOptionalComplexUnion`. Also added java doc to explain the semantics to make ti clearer.

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
```diff
@@ -96,7 +96,7 @@ public Type record(Schema record, List<String> names, List<Type> fieldTypes) {
       Object defaultValue = field.hasDefaultValue() && !(field.defaultVal() instanceof JsonProperties.Null) ?
           field.defaultVal() : null;
 
-      if (AvroSchemaUtil.isOptionSchema(field.schema())) {
+      if (AvroSchemaUtil.isOptionSchema(field.schema()) || AvroSchemaUtil.isNonOptionalUnionOptional(field.schema())) {
```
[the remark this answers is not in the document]
funcheetah: Discussed offline. We need to think the semantics of union further and improve it in a followup PR.

wmoustafa (approved):

PR #68, opened 2021-05-06
Change error handling to not delete data from HDFS when hive metaStore has been updated

ZihanLi58:
Issue: we encounter the issue that when shutdown happens during commit process, it leaves table in a dirty state (int the following commit/query, it ends up with table pointing to a non exist location). 
Solution: Change the error handling logic, when meet issue, will first check whether metadata store has been updated then decide whether to clean metadata file. If client closed or meta store unavailable, then not delete metadata file to avoid this situation and rely on the  process of "[Remove orphan files](https://iceberg.apache.org/maintenance/)" to remove untracked metadata file

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveMetadataPreservingTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveMetadataPreservingTableOperations.java
@@ -215,11 +215,30 @@ public class HiveMetadataPreservingTableOperations extends HiveTableOperations {
       throw new RuntimeException("Interrupted during commit", e);
 
     } finally {
-      if (threw) {
-        // if anything went wrong, clean up the uncommitted metadata file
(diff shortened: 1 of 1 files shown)


ZihanLi58:
@wmoustafa can you help take a look at this change? Thanks!

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveMetadataPreservingTableOperations.java
```diff
@@ -215,11 +215,30 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
       throw new RuntimeException("Interrupted during commit", e);
 
     } finally {
-      if (threw) {
-        // if anything went wrong, clean up the uncommitted metadata file
+      if (threw && !metadataUpdatedSuccessfully(newMetadataLocation)) {
```
wmoustafa, line 218: Is `threw` still required?

wmoustafa (approved):
Thanks @ZihanLi58 for this improvement.

PR #70, opened 2021-05-13
Default Value Support Prototype (not for review)

shenodaguirguis:
Sharing the prototype code for reference while reviewing the PRs, and for records.
In this prototype, [issue #2039 ](https://github.com/apache/iceberg/issues/2039) was reproduced via unit test, and the applied changes confirmed it fixes it and added other tests that default value is being used/returned whenever the field with default value is not manifested. This was tested for primitive as well as complex data types.

@wmoustafa


shenodaguirguis:
complete default-value support for AVRO PR: https://github.com/linkedin/iceberg/pull/72

PR #75, opened 2021-06-18
 [#2039] Support default value semantic for AVRO

shenodaguirguis:
Supporting default value semantics for AVRO tables/datasets.
cherry picked from master branch commits:

[[#2039] Support Default Value Semantis: use default values reading AV…](https://github.com/linkedin/iceberg/commit/b81ab9ed0c83cc20a29850a180ee7e56530a961b)

[[#2039] Support default value semantics: Schema Mapping changes](https://github.com/linkedin/iceberg/commit/c88ec3f2352789a213f3d409cec3ec36238d0ef0)
and 
[[#2039] Support default value semantics: API changes](https://github.com/linkedin/iceberg/commit/c18f4c4858deeb0e3b2a2831d18bc8dc163e22dd)

With these commits, on spark-shell one can select missing columns with non-null default values and get the default value in the results:

```
// required with non-null default
scala> spark.sql(s"select * from u_sguirgui.testi6").show
+----------+------------------+
| firstname |      lastname  |
+----------+------------------+
| Sam       |        Adam      |
| Adam      | default lastname |  <-- omitting the required col altogether
+----------+------------------+

// optional with default value
scala> DaliSpark.createDataFrame("dalids:///u_sguirgui.test5", Map(DaliSpark.PROJECT_COLS -> "firstname, lastname")).show
+----------+------------------+
| firstname|        lastname |
+----------+------------------+
| Sam      |    Adam         |
| Adam     | null.           |  <-- using the null option
| Adam.    | default lastname|  <-- omitting the col altogether
+----------+------------------+
```

PR #76, opened 2021-06-22
Support default value read for ORC format in spark

rzhang10:
This PR supports default value read for ORC format for Spark, based on the previous default value semantics support in #75.

During schema evolution, when there's an added field with the default value (regardless of required or optional field), the reader should return the default value for the field when reading the data files written with the previous schema (that doesn't contain this field).

When there's no schema evolution (i.e. when the field exists in the data file itself) and there's a default value for the field, the value written in the file should be read instead of the default value.

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -270,6 +270,9 @@ public final class ORCSchemaUtil {
           // Using suffix _r to avoid potential underlying issues in ORC reader
           // with reused column names between ORC and Iceberg;
           // e.g. renaming column c -> d and adding new column d
+          if (mapping.get(nestedField.fieldId()) == null && nestedField.hasDefaultValue()) {
+            continue;
+          }
           String name = Optional.ofNullable(mapping.get(nestedField.fieldId()))
               .map(OrcField::name)
               .orElseGet(() -> nestedField.name() + "_r" + nestedField.fieldId());
@@ -387,7 +390,7 @@ public final class ORCSchemaUtil {
         .map(Integer::parseInt);
   }
 
-  static int fieldId(TypeDescription orcType) {
+  public static int fieldId(TypeDescription orcType) {
     String idStr = orcType.getAttributeValue(ICEBERG_ID_ATTRIBUTE);
     Preconditions.checkNotNull(idStr, "Missing expected '%s' property", ICEBERG_ID_ATTRIBUTE);
     return Integer.parseInt(idStr);
--- a/spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.spark;
+
+import org.apache.iceberg.orc.ORCSchemaUtil;
+import org.apache.iceberg.orc.OrcSchemaWithTypeVisitor;
+import org.apache.iceberg.orc.OrcValueReader;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.types.Types;
+import org.apache.orc.TypeDescription;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public abstract class OrcSchemaWithTypeVisitorSpark<T> extends OrcSchemaWithTypeVisitor<T> {
+    protected final Map<Integer, Object> idToConstant;
+
+    protected OrcSchemaWithTypeVisitorSpark(Map<Integer, ?> idToConstant) {
+        this.idToConstant = new HashMap<>();
+        this.idToConstant.putAll(idToConstant);
+    }
+
+    @Override
+    protected T visitRecord(
+            Types.StructType struct, TypeDescription record, OrcSchemaWithTypeVisitor<T> visitor) {
+        Preconditions.checkState(
+                checkIcebergAndOrcSchemaAlignment(struct, record),
+                "Iceberg schema and ORC schema doesn't align, please call ORCSchemaUtil.buildOrcProjection" +
+                        "to get an aligned ORC schema first!"
+        );
+        List<Types.NestedField> iFields = struct.fields();
+        List<TypeDescription> fields = record.getChildren();
+        List<String> names = record.getFieldNames();
+        List<T> results = Lists.newArrayListWithExpectedSize(fields.size());
+
+        for (int i = 0, j = 0; i < iFields.size(); i++) {
+            Types.NestedField iField = iFields.get(i);
+            TypeDescription field = j < fields.size() ? fields.get(j) : null;
+            if (field == null || (iField.fieldId() != ORCSchemaUtil.fieldId(field))) {
+                idToConstant.put(iField.fieldId(), iField.getDefaultValue());
+            } else {
+                results.add(visit(iField.type(), field, visitor));
+                j++;
+            }
+        }
+        return visitor.record(struct, record, names, results);
+    }
+
+    private static boolean checkIcebergAndOrcSchemaAlignment(Types.StructType struct, TypeDescription record) {
+        List<Integer> icebergIDList = struct.fields().stream().map(Types.NestedField::fieldId).collect(Collectors.toList());
+        List<Integer> orcIDList = record.getChildren().stream().map(ORCSchemaUtil::fieldId).collect(Collectors.toList());
+
+        // icebergIDList should be a superset of orcIDList, and the overlapping ids should appear
+        // in the same order in these 2 lists
+        return checkTwoListAlignmentHelper(icebergIDList, orcIDList);
+    }
+
+    private static boolean checkTwoListAlignmentHelper(List<Integer> list1, List<Integer> list2) {
+        if (list1.size() < list2.size()) {
+            return false;
+        }
+
+        for (int i = 0, j = 0; j < list2.size(); j++) {
+            if (i >= list1.size()) {
+                return false;
+            }
+            while (!list1.get(i).equals(list2.get(j))) {
+                i++;
+                if (i >= list1.size()) {
+                    return false;
+                }
+            }
+            i++;
+        }
+        return true;
+    }
+}
--- a/spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
@@ -28,6 +28,7 @@ import org.apache.iceberg.orc.ORCSchemaUtil;
 import org.apache.iceberg.orc.OrcSchemaWithTypeVisitor;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.spark.source.BaseDataReader;
 import org.apache.iceberg.types.Types;
 import org.apache.orc.TypeDescription;
 
@@ -48,7 +49,7 @@ public abstract class OrcSchemaWithTypeVisitorSpark<T> extends OrcSchemaWithType
   protected T visitRecord(
           Types.StructType struct, TypeDescription record, OrcSchemaWithTypeVisitor<T> visitor) {
     Preconditions.checkState(
-            checkIcebergAndOrcSchemaAlignment(struct, record),
+            icebergFiledIdsContainOrcFieldIdsInOrder(struct, record),
             "Iceberg schema and ORC schema doesn't align, please call ORCSchemaUtil.buildOrcProjection" +
                     "to get an aligned ORC schema first!"
     );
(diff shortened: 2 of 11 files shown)


spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    protected T visitRecord(
+            Types.StructType struct, TypeDescription record, OrcSchemaWithTypeVisitor<T> visitor) {
+        Preconditions.checkState(
+                checkIcebergAndOrcSchemaAlignment(struct, record),
+                "Iceberg schema and ORC schema doesn't align, please call ORCSchemaUtil.buildOrcProjection" +
+                        "to get an aligned ORC schema first!"
+        );
+        List<Types.NestedField> iFields = struct.fields();
+        List<TypeDescription> fields = record.getChildren();
```
shenodaguirguis: rename to `structFields` and `recordFields`?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  protected T visitRecord(
+          Types.StructType struct, TypeDescription record, OrcSchemaWithTypeVisitor<T> visitor) {
+    Preconditions.checkState(
+            checkIcebergAndOrcSchemaAlignment(struct, record),
+            "Iceberg schema and ORC schema doesn't align, please call ORCSchemaUtil.buildOrcProjection" +
+                    "to get an aligned ORC schema first!"
+    );
+    List<Types.NestedField> iFields = struct.fields();
```
shenodaguirguis, line 56: iFields ?! should we rename to `icebergFields`?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+    return visitor.record(struct, record, names, results);
+  }
+
+  private static boolean checkIcebergAndOrcSchemaAlignment(Types.StructType struct, TypeDescription record) {
+    List<Integer> icebergIDList = struct.fields().stream().map(Types.NestedField::fieldId).collect(Collectors.toList());
+    List<Integer> orcIDList = record.getChildren().stream().map(ORCSchemaUtil::fieldId).collect(Collectors.toList());
+
+    // icebergIDList should be a superset of orcIDList, and the overlapping ids should appear
```
shenodaguirguis: this comment should be the java doc of this function

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+      } else {
+        results.add(visit(iField.type(), field, visitor));
+        j++;
+      }
+    }
+    return visitor.record(struct, record, names, results);
+  }
+
+  private static boolean checkIcebergAndOrcSchemaAlignment(Types.StructType struct, TypeDescription record) {
```
shenodaguirguis: rename to `icebergFiledIdsContainOrcFieldIdsInOrder` ?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+    List<Integer> icebergIDList = struct.fields().stream().map(Types.NestedField::fieldId).collect(Collectors.toList());
+    List<Integer> orcIDList = record.getChildren().stream().map(ORCSchemaUtil::fieldId).collect(Collectors.toList());
+
+    // icebergIDList should be a superset of orcIDList, and the overlapping ids should appear
+    // in the same order in these 2 lists
+    return checkTwoListAlignmentHelper(icebergIDList, orcIDList);
+  }
+
+  private static boolean checkTwoListAlignmentHelper(List<Integer> list1, List<Integer> list2) {
```
shenodaguirguis: rename to `containsInOrder` ?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+    if (list1.size() < list2.size()) {
+      return false;
+    }
+
+    for (int i = 0, j = 0; j < list2.size(); j++) {
+      if (i >= list1.size()) {
```
shenodaguirguis, line 107: hoist this check to the for loop condition?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.stream.Collectors;
+import org.apache.iceberg.MetadataColumns;
+import org.apache.iceberg.orc.ORCSchemaUtil;
+import org.apache.iceberg.orc.OrcSchemaWithTypeVisitor;
+import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.types.Types;
+import org.apache.orc.TypeDescription;
+
+public abstract class OrcSchemaWithTypeVisitorSpark<T> extends OrcSchemaWithTypeVisitor<T> {
```
shenodaguirguis, line 35: why do we need a new visitor that only modifies the struct visitor (as opposed to modify the parent/existing one)? Are there cases where we would still want to use the parent visitor (i.e., not handling default values)?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+    for (int i = 0, j = 0; i < iFields.size(); i++) {
+      Types.NestedField iField = iFields.get(i);
+      TypeDescription field = j < fields.size() ? fields.get(j) : null;
+      if (field == null || (iField.fieldId() != ORCSchemaUtil.fieldId(field))) {
+        // there are 3 cases where we need to use idToConstant for an iField
+        // 1. The field is MetadataColumns.ROW_POSITION, we build a RowPositionReader
+        // 2. The field is a partition column, we build a ConstantReader
+        // 3. The field should be read using the default value, where we build a ConstantReader
+        // Here we should only need to update idToConstant when it's the 3rd case,
+        // because the first 2 cases have been handled by logic elsewhere.
```
shenodaguirguis: it would be helpful to specify the `elsewhere`

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
```diff
@@ -0,0 +1,110 @@
...
+        if (!iField.equals(MetadataColumns.ROW_POSITION) &&
```
shenodaguirguis, line 71: shouldn't we use `MetadataColumns.metadataFieldIds().contains` similar to AVRO case?

spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ConstantColumnVector.java
```diff
@@ -109,6 +109,9 @@ public Decimal getDecimal(int rowId, int precision, int scale) {
 
   @Override
   public UTF8String getUTF8String(int rowId) {
+    if (constant instanceof String) {
```
shenodaguirguis: interesting! why is this special handling needed?

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcReaderForFieldsWithDefaultValue.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.apache.iceberg.spark.data.TestHelpers.assertEquals;
+
+
+public class TestSparkOrcReaderForFieldsWithDefaultValue {
+
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  @Test
+  public void testOrcDefaultValues() throws IOException {
```
shenodaguirguis, line 61: can we have separate tests for different readers, and can we also test different field types, as we did with avro?

shenodaguirguis (commented):
Thanks for this patch @rzhang10 ! Left comments that are mostly related to code readability. The major comment would be regarding the need to add a new visitor.

spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ConstantColumnVector.java
shardulm94: Instead of special handling here, I would suggest using `BaseDataReader::convertConstant` to convert constants from the Iceberg representation to Spark representation beforehand. This is what we do for other constants in the `idToConstant` map.

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
rzhang10: The existing codebase have being using names like `iSchema` and `iField` to refer to iceberg related interfaces.

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
rzhang10: My rationale is that, the parent is a more generic ORC visitor, whereas the ones I intend to modify is only related to spark behavior. I want the overridden behavior to only affect the scope of spark, but leave the generic visitor for other future prospective extensions, that's why I structure my code this way.

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
rzhang10: There are only 2 `MetadataColumns`, one is row position and another is file path, I think we only need to consider row position here since we want to build a row position reader to construct a row position column at runtime.

spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ConstantColumnVector.java
rzhang10: Yeah it's weird, it seems spark vectorized API expect a string type to be `UTF8String`, so when we have a default value object of String type, we need to convert explicitly to `UTF8String`

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
shenodaguirguis: fair

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
shenodaguirguis: I see. The goal, however, is make Iceberg support default values for AVRO and ORC formats, not only for Spark. The changes for AVRO are along these lines.

spark/src/main/java/org/apache/iceberg/spark/data/vectorized/ConstantColumnVector.java
shenodaguirguis: I see. Thanks for the explanation!

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcReaderForFieldsWithDefaultValue.java
funcheetah: +1 on more tests for more complex schema

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -270,6 +270,9 @@ private static TypeDescription buildOrcProjection(Integer fieldId, Type type, bo
           // Using suffix _r to avoid potential underlying issues in ORC reader
           // with reused column names between ORC and Iceberg;
           // e.g. renaming column c -> d and adding new column d
+          if (mapping.get(nestedField.fieldId()) == null && nestedField.hasDefaultValue()) {
```
funcheetah, line 273: Could you explain the logic here?

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
funcheetah: @rzhang10 , is the logic implemented in a way that only supports Spark behavior?

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
rzhang10: The default value and the corresponding schema are only kept at iceberg level, we shouldn't include such field in the projected ORC schema, cuz the field doesn't exist in the physical ORC data file anyways. The default value will be handled by iceberg itself via the `ConstantReader`

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
rzhang10: Actually, it is not just for spark, it supports the default value for ORC regardless of the engine, I just don't feel like changing the original visitor code to minimize the potential impact, but we could as well just modify the `OrcSchemaWithTypeVisitor` code itself without using an extended subclass.

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
rzhang10: I think it's fine to leave it like this since other iceberg code use the `i` vs non `i` pattern

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
shenodaguirguis: if the support for default values is not limited by the engine type, then it is fine from that perspective to use a subclass. However, for code consistency, we need to use the same methodology for both union types and default values, for ORC  (i.e.,  either both modify or both extend the visitor). I will leave it to you @rzhang10  and @funcheetah  to decide which way makes more sense).

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcReaderForFieldsWithDefaultValue.java
rzhang10: Added nested data type support in the new commit and also tests.

spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java
rzhang10: Actually, after my new patch (which uses the `BaseDataReader.convertConstant` to convert iceberg constant to spark constant), the visitor has become Spark-specific, since its visit method directly call `convertConstant` which is spark specific conversion logic. So this can adds to the reason why I stick with my current class inheritance structure.

shenodaguirguis: that is fine for now. Can you please create an issue to support default values for ORC, for non-spark readers?

shenodaguirguis (approved):
Thanks for enabling default values for Spark/ORC @rzhang10!

funcheetah (approved):
This PR looks good to me. Thanks @rzhang10 for the great work!

funcheetah (approved):

PR #77, opened 2021-06-25
Support Date and Timestamp hive type partition columns in LegacyHiveT…

rzhang10:
Previously, `LegacyHiveTableScan` only supports hive partition columns of string type, this patch leverage the `Binder` to augment the expressions (literals) with their corresponding types information and feed it to `ExpressionToPartitionFilterString`, thus has the power to handle different partition column types such as `Timestamp` and `Date`.

After this patch, iceberg can support reading partitioned tables in our existing hive metastore which has `date` and `timestamp` typed partition columns.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
@@ -19,6 +19,10 @@
 
 package org.apache.iceberg.hive.legacy;
 
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -29,6 +33,7 @@ import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.Literal;
 import org.apache.iceberg.expressions.UnboundPredicate;
 import org.apache.iceberg.expressions.UnboundTerm;
+import org.apache.iceberg.types.Type;
 
 
 class HiveExpressions {
@@ -238,6 +243,7 @@ class HiveExpressions {
   }
 
   private static class ExpressionToPartitionFilterString extends ExpressionVisitors.ExpressionVisitor<String> {
+    private static final OffsetDateTime EPOCH = Instant.ofEpochSecond(0).atOffset(ZoneOffset.UTC);
     private static final ExpressionToPartitionFilterString INSTANCE = new ExpressionToPartitionFilterString();
 
     private ExpressionToPartitionFilterString() {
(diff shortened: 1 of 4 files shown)


hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
```diff
@@ -315,8 +324,17 @@ private String getOperationString(Expression.Operation op) {
       }
     }
 
-    private <T> String getLiteralValue(Literal<T> lit) {
+    private <T> String getLiteralValue(Literal<T> lit, Type type) {
       Object value = lit.value();
+      switch (type.typeId()) {
```
shenodaguirguis, line 329: should we `assert` in `default` that `type` is primitive type?

hive-metastore/src/test/java/org/apache/iceberg/hive/legacy/TestHiveExpressions.java
```diff
@@ -118,10 +117,4 @@ public void testSimplifyRemoveNonPartitionColumnsWithinNot2() {
     Expression expected = alwaysTrue();
     Assert.assertEquals(expected.toString(), simplifyPartitionFilter(input, ImmutableSet.of("pcol")).toString());
   }
-
-  @Test
-  public void testToPartitionFilterStringEscapeStringLiterals() {
```
shenodaguirguis, line 123: why remove?

hive-metastore/src/test/java/org/apache/iceberg/hive/legacy/TestHiveExpressions.java
```diff
@@ -36,7 +36,6 @@
 import static org.apache.iceberg.expressions.Expressions.notNull;
 import static org.apache.iceberg.expressions.Expressions.or;
 import static org.apache.iceberg.hive.legacy.HiveExpressions.simplifyPartitionFilter;
-import static org.apache.iceberg.hive.legacy.HiveExpressions.toPartitionFilterString;
 
 
 public class TestHiveExpressions {
```
shenodaguirguis, line 41: can we add tests for DATE and TIMESTAMP (and other datatypes we should support)?

shenodaguirguis (commented):
Thanks for this patch @rzhang10! 
Couple minor comments.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/HiveExpressions.java
rzhang10: I'm not completely sure what are the allowed types in hive for the partition columns, but in its internal API it always uses a string value in the end for a partition value. So I think it's fine we leave the code like this.

shenodaguirguis: ok

rzhang10:
Please review #78 , which is a superset of this PR.

PR #78, opened 2021-07-01
Support hive non string partition cols

rzhang10:
Previously, `LegacyHiveTableScan` only supports hive partition columns of string type.
This patch leverages the `Binder` to augment the expressions (literals) with their corresponding types of information and feed it to `ExpressionToPartitionFilterString`, thus handling different partition column types such as `Timestamp` and `Date`.
It also adds a fall-back filtering strategy when the HMS `listPartitionsByFilter` call failed because of non-string type partition columns, by first listing all partitions from HMS, and then do the filtering ourselves by evaluating each partition against the partitions filter `BoundExpression`.

After this patch, Iceberg can support reading partitioned tables in our existing hive metastore which has `date`, `timestamp`, `int`, `varchar(n)`, `char(n)` typed partition columns.

This PR covers all the features introduced in #77, so will close #77 now.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/expressions/And.java
+++ b/api/src/main/java/org/apache/iceberg/expressions/And.java
@@ -60,7 +60,7 @@ public class And implements Expression, Bound<Boolean> {
     Preconditions.checkNotNull(left, "Left expression cannot be null.");
     Preconditions.checkNotNull(right, "Right expression cannot be null.");
     if (!(left instanceof Bound) || !(right instanceof Bound)) {
-      throw new IllegalStateException("Bound predicate not expected");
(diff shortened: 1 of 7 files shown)


shenodaguirguis:
Thanks for the quick turn around @rzhang10! 
Is this strategy (to fall back and evaluate expression ourselves) the same as what spark native readers do?

rzhang10:
> Thanks for the quick turn around @rzhang10!
> Is this strategy (to fall back and evaluate expression ourselves) the same as what spark native readers do?

Correct, Spark native reader path has this behavior.

shenodaguirguis (approved):
LGTM! Thanks @rzhang10 !

funcheetah (approved):

PR #80, opened 2021-07-27
Support avro.schema.literal/hive union types in Hive legacy table to …

rzhang10:
…Iceberg conversion

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/MergeHiveSchemaWithAvro.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/MergeHiveSchemaWithAvro.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.hive.legacy;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.avro.JsonProperties;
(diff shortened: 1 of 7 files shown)


funcheetah (approved):
Thanks @rzhang10 for this patch. LGTM.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/MergeHiveSchemaWithAvro.java
```diff
@@ -177,6 +190,15 @@ public Schema listElementPartner(Schema partner) {
       return (schema.getType() == Schema.Type.ARRAY) ? schema.getElementType() : null;
     }
 
+    @Override
+    public Schema unionObjectPartner(Schema partner, int ordinal) {
+      if (partner.getType() != Schema.Type.UNION) {
+        return null;
+      }
+      Schema schema = AvroSchemaUtil.discardNullFromUnionIfExist(partner);
```
shenodaguirguis, line 198: just curious, why do we need to drop the null? Do they get dropped somewhere else along the way? If so, do we know why & where? Is there e another util to use instead of adding`AvroSchemaUtil.discardNullFromUnionIfExist`?

rzhang10: Because in the partner visitor, Hive schema is visited along with the Avro schema, Hive union's schema won't contain null, so we want to discard it temporarily for the benefit of the subsequential nested visiting. Later I added the null back to the avro schema in `AvroSchemaUtil.nullExistInUnion(partner)` condition call.

shenodaguirguis: that makes sense. Thanks Raymond!

PR #81, opened 2021-09-02
Fix ORC schema visitors to support reading ORC files with deeply nest…

rzhang10:
…ed union type schema

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaVisitor.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaVisitor.java
@@ -49,15 +49,14 @@ public abstract class OrcSchemaVisitor<T> {
       case UNION:
         List<TypeDescription> types = schema.getChildren();
         List<T> options = Lists.newArrayListWithExpectedSize(types.size());
-        for (TypeDescription type : types) {
-          visitor.beforeUnionOption(type);
+        for (int i = 0; i < types.size(); i++) {
+          visitor.beforeUnionOption(types.get(i), i);
           try {
-            options.add(visit(type, visitor));
+            options.add(visit(types.get(i), visitor));
           } finally {
-            visitor.afterUnionOption(type);
+            visitor.afterUnionOption(types.get(i), i);
           }
         }
-
         return visitor.union(schema, options);
 
       case LIST:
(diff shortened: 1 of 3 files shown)


orc/src/main/java/org/apache/iceberg/orc/OrcSchemaVisitor.java
```diff
@@ -123,8 +122,8 @@
     return visitor.record(record, names, visitFields(fields, names, visitor));
   }
 
-  public String optionName() {
-    return "_option";
+  public String optionName(int ordinal) {
+    return "tag_" + ordinal;
```
funcheetah, line 126: Could you give an example how this is used?

funcheetah:
Thanks @rzhang10 for the quick fix.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
```diff
@@ -215,4 +216,83 @@ public void testSingleComponentUnion() throws IOException {
...
+        innerInt1.fillWithNulls();
+        innerString3.fillWithNulls();
+      }
+    }
+    if (batch.size != 0) {
+      writer.addRowBatch(batch);
+      batch.reset();
+    }
+    writer.close();
+
```
funcheetah, line 286: Could you also add a test for `VectorizedSparkOrcReaders`?

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaVisitor.java
rzhang10: It's used by the `ApplyNameMapping` visitor to correctly recognize the union converted struct field names, for example sth like `c1.tag_1.c2` and assign the correct ID to such field.

shenodaguirguis: I didn't check the ApplyNameMapping visitor, but can you verify that using same ordinal across unions - i.e., (1, "tag_0") and again (5, "tag_0") as in your test case -  could lead to any issues?

shenodaguirguis (approved):
LGTM, left one question to verify

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
rzhang10: Added

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaVisitor.java
rzhang10: It won't, the (1, tag_0) and (5, tag_0) are two fields at different nested levels inside the schema so there's no ambiguity.

funcheetah (approved):

PR #82, opened 2021-09-14
Move to Avro 1.10.2

shenodaguirguis:
Iceberg currently uses Avro 1.9, which throws an Exception for Malformed/wrong-typed default values. Now that we added support for default-values in Iceberg, for ORC and AVRO formats, this might cause trouble with ecosystems that uses AVRO 1.7, which only warns about malformed/wrong-typed defaults values and silently sets no-default-value. Avro 1.10.2 provides an API that gives the caller the option to skip validating the default value. 
Note: it will still fail at the Iceberg schema level, given wronged-typed default value, which will give the caller the meaningful error to correct their schema.

### DIFF
--- a/core/src/test/java/org/apache/iceberg/TestManifestListVersions.java
+++ b/core/src/test/java/org/apache/iceberg/TestManifestListVersions.java
@@ -27,6 +27,7 @@ import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecordBuilder;
 import org.apache.iceberg.avro.Avro;
 import org.apache.iceberg.avro.AvroSchemaUtil;
+import org.apache.iceberg.avro.TestReadProjection;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.FileAppender;
 import org.apache.iceberg.io.InputFile;
@@ -132,9 +133,9 @@ public class TestManifestListVersions {
     Assert.assertEquals("Added rows count", ADDED_ROWS, (long) generic.get("added_rows_count"));
     Assert.assertEquals("Existing rows count", EXISTING_ROWS, (long) generic.get("existing_rows_count"));
     Assert.assertEquals("Deleted rows count", DELETED_ROWS, (long) generic.get("deleted_rows_count"));
-    Assert.assertNull("Content", generic.get(ManifestFile.MANIFEST_CONTENT.name()));
-    Assert.assertNull("Sequence number", generic.get(ManifestFile.SEQUENCE_NUMBER.name()));
-    Assert.assertNull("Min sequence number", generic.get(ManifestFile.MIN_SEQUENCE_NUMBER.name()));
+    TestReadProjection.assertNotProjected("Content", generic, ManifestFile.MANIFEST_CONTENT.name());
+    TestReadProjection.assertNotProjected("Sequence number", generic, ManifestFile.SEQUENCE_NUMBER.name());
+    TestReadProjection.assertNotProjected("Min sequence number", generic, ManifestFile.MIN_SEQUENCE_NUMBER.name());
   }
 
   @Test
@@ -154,9 +155,9 @@ public class TestManifestListVersions {
     Assert.assertEquals("Added rows count", ADDED_ROWS, (long) generic.get("added_rows_count"));
     Assert.assertEquals("Existing rows count", EXISTING_ROWS, (long) generic.get("existing_rows_count"));
     Assert.assertEquals("Deleted rows count", DELETED_ROWS, (long) generic.get("deleted_rows_count"));
-    Assert.assertNull("Content", generic.get(ManifestFile.MANIFEST_CONTENT.name()));
-    Assert.assertNull("Sequence number", generic.get(ManifestFile.SEQUENCE_NUMBER.name()));
-    Assert.assertNull("Min sequence number", generic.get(ManifestFile.MIN_SEQUENCE_NUMBER.name()));
+    TestReadProjection.assertNotProjected("Content", generic, ManifestFile.MANIFEST_CONTENT.name());
+    TestReadProjection.assertNotProjected("Sequence number", generic, ManifestFile.SEQUENCE_NUMBER.name());
+    TestReadProjection.assertNotProjected("Min sequence number", generic, ManifestFile.MIN_SEQUENCE_NUMBER.name());
   }
 
   @Test
--- a/core/src/test/java/org/apache/iceberg/avro/TestReadProjection.java
+++ b/core/src/test/java/org/apache/iceberg/avro/TestReadProjection.java
@@ -22,6 +22,7 @@ package org.apache.iceberg.avro;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.generic.GenericData.Record;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableList;
@@ -148,7 +149,7 @@ public abstract class TestReadProjection {
     );
 
     Record projected = writeAndRead("basic_projection_id", writeSchema, idOnly, record);
-    Assert.assertNull("Should not project data", projected.get("data"));
+    assertNotProjected("should not project data", projected, "data");
     Assert.assertEquals("Should contain the correct id value", 34L, (long) projected.get("id"));
 
     Schema dataOnly = new Schema(
@@ -157,7 +158,7 @@ public abstract class TestReadProjection {
 
     projected = writeAndRead("basic_projection_data", writeSchema, dataOnly, record);
 
-    Assert.assertNull("Should not project id", projected.get("id"));
+    assertNotProjected("Should not project id", projected, "id");
     int cmp = Comparators.charSequences()
         .compare("test", (CharSequence) projected.get("data"));
     Assert.assertTrue("Should contain the correct data value", cmp == 0);
@@ -210,9 +211,8 @@ public abstract class TestReadProjection {
     );
 
     Record projected = writeAndRead("id_only", writeSchema, idOnly, record);
-    Record projectedLocation = (Record) projected.get("location");
     Assert.assertEquals("Should contain the correct id value", 34L, (long) projected.get("id"));
-    Assert.assertNull("Should not project location", projectedLocation);
+    assertNotProjected("should not project location", projected, "location");
 
     Schema latOnly = new Schema(
         Types.NestedField.optional(3, "location", Types.StructType.of(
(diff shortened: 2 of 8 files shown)


shenodaguirguis:
@rzhang10 FYI

core/src/test/java/org/apache/iceberg/TestManifestListVersions.java
```diff
@@ -132,9 +133,9 @@ public void testV1ForwardCompatibility() throws IOException {
     Assert.assertEquals("Added rows count", ADDED_ROWS, (long) generic.get("added_rows_count"));
     Assert.assertEquals("Existing rows count", EXISTING_ROWS, (long) generic.get("existing_rows_count"));
     Assert.assertEquals("Deleted rows count", DELETED_ROWS, (long) generic.get("deleted_rows_count"));
-    Assert.assertNull("Content", generic.get(ManifestFile.MANIFEST_CONTENT.name()));
-    Assert.assertNull("Sequence number", generic.get(ManifestFile.SEQUENCE_NUMBER.name()));
-    Assert.assertNull("Min sequence number", generic.get(ManifestFile.MIN_SEQUENCE_NUMBER.name()));
+    TestReadProjection.assertNotProjected("Content", generic, ManifestFile.MANIFEST_CONTENT.name());
+    TestReadProjection.assertNotProjected("Sequence number", generic, ManifestFile.SEQUENCE_NUMBER.name());
+    TestReadProjection.assertNotProjected("Min sequence number", generic, ManifestFile.MIN_SEQUENCE_NUMBER.name());
```
wmoustafa, line 138: Is this related to the Avro upgrade?

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,11 @@ private LegacyHiveTableUtils() {
   static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
     Map<String, String> props = getTableProperties(table);
     String schemaStr = props.get("avro.schema.literal");
-    org.apache.avro.Schema avroSchema = schemaStr != null ? new org.apache.avro.Schema.Parser().parse(schemaStr) : null;
+    // at linkedin, avro 1.7 has been used heavily where no default value validation took place, which
+    // led to having a lot of production schemata with malformed defaults. Therefore, we disable
+    // the default value validation
```
wmoustafa: You can just say `// Disable default value validation for backward compatibility with Avro 1.7`.

parquet/src/main/java/org/apache/iceberg/parquet/ParquetAvro.java
```diff
@@ -169,12 +170,16 @@ public String getLogicalTypeName() {
...
+      return new BigDecimal(new BigInteger(value.bytes()), dec.scale());
     }
 
     @Override
     public GenericFixed toFixed(BigDecimal value, Schema schema, LogicalType type) {
-      return super.toFixed(value, schema, decimalsByScale[((ParquetDecimal) type).scale()]);
+      ParquetDecimal dec = (ParquetDecimal) type;
+      Pair<Integer, Integer> key = new Pair<>(dec.precision(), dec.scale());
+      return super.toFixed(value, schema,
+          decimalsByScale.computeIfAbsent(key, k -> LogicalTypes.decimal(k.getFirst(), k.getSecond())));
```
wmoustafa, line 182: Are changes in this file related to the Avro upgrade? Do they exist in the upstream PR?

core/src/test/java/org/apache/iceberg/TestManifestListVersions.java
shenodaguirguis: yes related, and similar changes are in the upstream PR: avro throwing instead of returning null

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
shenodaguirguis: thx for the suggestion

parquet/src/main/java/org/apache/iceberg/parquet/ParquetAvro.java
shenodaguirguis: yes, this is backported from the upstream PR. It relates to the way avro handles scale/precision

rzhang10:
I see a comment in upstream #1648 https://github.com/apache/iceberg/pull/1648#issuecomment-805284872 , should we also consider using 1.10.1 instead of 1.10.2 ?

shenodaguirguis:
> I see a comment in upstream #1648 [apache/iceberg#1648 (comment)](https://github.com/apache/iceberg/pull/1648#issuecomment-805284872) , should we also consider using 1.10.1 instead of 1.10.2 ?

@rzhang10 the avro parser feature to disable defaults validation exists starting avro 1.10.2, and this PR is mainly to use it. And yes I saw that comment and spent significant time to fix the jackson issue (the changes in build.gradle, specifically `force 'com.fasterxml.jackson.core:jackson-databind:2.10.2'`). I also created jar and tested with spark-shell, so we should be good.

parquet/src/test/java/org/apache/iceberg/avro/TestReadProjection.java
```diff
@@ -513,17 +513,26 @@ public void testListOfStructsProjection() throws IOException {
...
-    Assert.assertNull("Should not project y", projectedP2.get("y"));
+    assertNotProjected("Should not project x", projectedP2, "x");
+    assertNotProjected("Should not project y", projectedP2, "y");
     Assert.assertEquals("Should project null z", null, projectedP2.get("z"));
   }
+
+  public static void assertNotProjected(String message, Record projected, String fieldName) {
+    try {
+      projected.get(fieldName);
+      Assert.fail(message);
```
rzhang10, line 533: I feel we don't need this line.

core/src/test/java/org/apache/iceberg/avro/TestReadProjection.java
```diff
@@ -513,17 +513,26 @@ public void testListOfStructsProjection() throws IOException {
...
-    Assert.assertNull("Should not project y", projectedP2.get("y"));
+    assertNotProjected("Should not project x", projectedP2, "x");
+    assertNotProjected("Should not project y", projectedP2, "y");
     Assert.assertEquals("Should project null z", null, projectedP2.get("z"));
   }
+
+  public static void assertNotProjected(String message, Record projected, String fieldName) {
+    try {
+      projected.get(fieldName);
+      Assert.fail(message);
```
rzhang10, line 533: I feel we don't need this line.

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
```diff
@@ -56,7 +56,9 @@ private LegacyHiveTableUtils() {
   static Schema getSchema(org.apache.hadoop.hive.metastore.api.Table table) {
     Map<String, String> props = getTableProperties(table);
     String schemaStr = props.get("avro.schema.literal");
-    org.apache.avro.Schema avroSchema = schemaStr != null ? new org.apache.avro.Schema.Parser().parse(schemaStr) : null;
+    // Disable default value validation for backward compatibility with Avro 1.7
+    org.apache.avro.Schema avroSchema =
+        schemaStr != null ? new org.apache.avro.Schema.Parser().setValidateDefaults(false).parse(schemaStr) : null;
```
rzhang10, line 61: I just have a general question here: I understand here we use `setValidateDefaults(false)` to bypass mal-formatted default values validation here.
Is it the case that in the later code logic, we will only fail when the subsequent readers try to read such mal-formatted default value fields, and code runs gracefully otherwise; or is it the case that the later logic like schema conversion: eg things like Avro schema -> iceberg schema conversion might still fail?
To me the correct behavior is the first, have we verified this?

parquet/src/test/java/org/apache/iceberg/avro/TestReadProjection.java
shenodaguirguis: this is to mimic the `Assert.assertNull` w/ error message that this API replaces.

core/src/test/java/org/apache/iceberg/avro/TestReadProjection.java
shenodaguirguis: same as above

wmoustafa (approved):

hive-metastore/src/main/java/org/apache/iceberg/hive/legacy/LegacyHiveTableUtils.java
funcheetah: +1 on this

shenodaguirguis: Good question! Yes, it will fail only when subsequent readers try to read such malformed default values because it is replaced with a "null" default when we skip default validations (I believe this is similar to avro 1.7 behavior). This is as opposed to avro 1.9 behavior which fails whenever reading the table, regardless if has data or not, and whether the data has the defaulted field filled or not (i.e., need to use the default value or not)

Below are different scenarios:

1) reading from empty table with malformed default: 

scala> val df = spark.sql("select * from u_sguirgui.test10")
[WARNING] Avro: Invalid default for field id: "one" not a "int"
df: org.apache.spark.sql.DataFrame = [id: int, address: string]

scala> df.show
+---+-------+
| id|address|
+---+-------+
+---+-------+

2) reading from table with malformed defaults, but rows populate all fields (does not need to  user the default value):
scala> spark.sql("select * from u_sguirgui.test10").show
21/09/21 05:09:47 WARN HiveConf: HiveConf of name hive.metastore.token.signature does not exist
Hive history file=/grid/hive_query_logs/sguirgui/hive_job_log_bd3f4c79-b990-4378-80d9-107944b1f034_1771310506.txt
[WARNING] Avro: Invalid default for field id: "one" not a "int"
21/09/21 05:09:49 WARN HiveConf: HiveConf of name hive.metastore.token.signature does not exist
21/09/21 05:09:49 WARN HiveConf: HiveConf of name hive.metastore.token.signature does not exist
+---+---------+                                                                 
| id|  address|
+---+---------+
|  2|address 2|
+---+---------+


3) reading from table with malformed defaults, and some rows has missing defaulted field:

scala> spark.sql("select * from u_sguirgui.test10").show
[WARNING] Avro: Invalid default for field id: "one" not a "int"
[Stage 1:>                                                          (0 + 1) / 1]21/09/21 05:17:26 WARN TaskSetManager: Lost task 0.0 in stage 1.0 (TID 1, ltx1-hcl9074.grid.linkedin.com, executor 2): java.lang.IllegalArgumentException: Missing required field that has no default value: expectedField: 0: id: required int, avroField: null, record: {"type":"record","name":"test11","namespace":"u_sguirgui","fields":[{"name":"address","type":["null","string"],"default":null,"field-id":1}]}
	at dali2_shaded.org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:441)
	at dali2_shaded.org.apache.iceberg.avro.BuildAvroProjection.record(BuildAvroProjection.java:112)
	at dali2_shaded.org.apache.iceberg.avro.BuildAvroProjection.record(BuildAvroProjection.java:42)
	at dali2_shaded.org.apache.iceberg.avro.AvroCustomOrderSchemaVisitor.visit(AvroCustomOrderSchemaVisitor.java:51)
	at dali2_shaded.org.apache.iceberg.avro.AvroSchemaUtil.buildAvroProjection(AvroSchemaUtil.java:106)

PR #83, opened 2021-10-18
Fix spark avro reader reading union schema data

rzhang10:
This PR fixes 2 implementation issues in #73 

1. The previous implementation doesn't handle return values correctly when there is a `null` element in the union, like `[null, int, string]`, it returned a struct data of `[null, 1, null` instead of just `[1, null]`, this PR fixes it.
2. The previous implementation doesn't decorate the internal Avro schema with correct mapped field-ids, when there's nested complex data types inside the union, because the union to struct conversion create another layer of a struct that contains several fields, each of which has a new field name and brings in a new namespace, this PR is fixing that.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaVisitor.java
@@ -52,8 +52,21 @@ public abstract class AvroSchemaVisitor<T> {
       case UNION:
         List<Schema> types = schema.getTypes();
         List<T> options = Lists.newArrayListWithExpectedSize(types.size());
-        for (Schema type : types) {
-          options.add(visit(type, visitor));
+        if (AvroSchemaUtil.isOptionSchema(schema)) {
+          for (Schema type : types) {
+            options.add(visit(type, visitor));
+          }
+        } else {
+          int i = 0;
+          // complex union case
+          for (Schema type : types) {
+            if (type.getType() != Schema.Type.NULL) {
+              options.add(visitWithName("tag_"+i, type, visitor));
+              i += 1;
+            } else {
+              options.add(visit(type, visitor));
+            }
+          }
         }
         return visitor.union(schema, options);
 
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaVisitor.java
@@ -57,12 +57,12 @@ public abstract class AvroSchemaVisitor<T> {
             options.add(visit(type, visitor));
           }
         } else {
-          int i = 0;
           // complex union case
+          int idx = 0;
           for (Schema type : types) {
             if (type.getType() != Schema.Type.NULL) {
-              options.add(visitWithName("tag_"+i, type, visitor));
-              i += 1;
+              options.add(visitWithName("tag_" + idx, type, visitor));
+              idx += 1;
             } else {
               options.add(visit(type, visitor));
             }
--- a/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
+++ b/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
@@ -125,7 +125,6 @@ class PruneColumns extends AvroSchemaVisitor<Schema> {
       return null;
     } else {
       // Complex union case
-//      return Schema.createUnion(options);
       return copyUnion(union, options);
     }
   }
--- a/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
+++ b/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
@@ -299,10 +299,13 @@ class PruneColumns extends AvroSchemaVisitor<Schema> {
     return AvroSchemaUtil.isOptionSchema(schema) && schema.getTypes().get(0).getType() != Schema.Type.NULL;
   }
 
+  // for primitive types, the visitResult will be null, we want to reuse the primitive types from the original
+  // schema, while for nested types, we want to use the visitResult because they have content from the previous
+  // recursive calls.
   private static Schema copyUnion(Schema record, List<Schema> visitResults) {
     List<Schema> alts = Lists.newArrayListWithExpectedSize(visitResults.size());
     for (int i = 0; i < visitResults.size(); i++) {
-      if (visitResults.get(0) == null) {
+      if (visitResults.get(i) == null) {
         alts.add(record.getTypes().get(i));
       } else {
         alts.add(visitResults.get(i));
--- a/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
+++ b/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
@@ -125,9 +125,11 @@ class PruneColumns extends AvroSchemaVisitor<Schema> {
       return null;
     } else {
       // Complex union case
-      return union;
+//      return Schema.createUnion(options);
+      return copyUnion(union, options);
     }
   }
+
 
   @Override
   @SuppressWarnings("checkstyle:CyclomaticComplexity")
@@ -296,5 +298,17 @@ class PruneColumns extends AvroSchemaVisitor<Schema> {
 
   private static boolean isOptionSchemaWithNonNullFirstOption(Schema schema) {
     return AvroSchemaUtil.isOptionSchema(schema) && schema.getTypes().get(0).getType() != Schema.Type.NULL;
+  }
+
+  private static Schema copyUnion(Schema record, List<Schema> visitResults) {
+    List<Schema> alts = Lists.newArrayListWithExpectedSize(visitResults.size());
+    for (int i = 0; i < visitResults.size(); i++) {
+      if (visitResults.get(0) == null) {
+        alts.add(record.getTypes().get(i));
+      } else {
+        alts.add(visitResults.get(i));
+      }
+    }
+    return Schema.createUnion(alts);
   }
 }
--- a/spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
@@ -28,7 +28,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-
 import org.apache.avro.Schema;
 import org.apache.avro.io.Decoder;
 import org.apache.avro.util.Utf8;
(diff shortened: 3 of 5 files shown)


spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -96,13 +101,14 @@ public void writeAndValidateOptionalComplexUnion() throws IOException {
...
-        .noDefault()
+        .nullDefault()
         .endRecord();
 
     GenericData.Record unionRecord1 = new GenericData.Record(avroSchema);
-    unionRecord1.put("unionCol", "StringType1");
+    unionRecord1.put("unionCol", "foo");
     GenericData.Record unionRecord2 = new GenericData.Record(avroSchema);
     unionRecord2.put("unionCol", 1);
+    List<GenericData.Record> expectedRows = new ArrayList<>();
```
ljfgem: 'expectedRows' is never used?

rzhang10: good find, I just removed it.

core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
```diff
@@ -297,4 +298,16 @@ private static Schema copyRecord(Schema record, List<Schema.Field> newFields) {
   private static boolean isOptionSchemaWithNonNullFirstOption(Schema schema) {
     return AvroSchemaUtil.isOptionSchema(schema) && schema.getTypes().get(0).getType() != Schema.Type.NULL;
   }
+
+  private static Schema copyUnion(Schema record, List<Schema> visitResults) {
+    List<Schema> alts = Lists.newArrayListWithExpectedSize(visitResults.size());
+    for (int i = 0; i < visitResults.size(); i++) {
+      if (visitResults.get(0) == null) {
```
ljfgem: Why do we use `visitResults.get(0) == null` instead of `visitResults.get(i) == null`?
And could you add comment for this method, preferably with an example.

rzhang10: Good find, this is a typo, it should be `get (i)`. This method is more of implementation detail, the reason is for primitive types, the `visitResult` will be null, so we want to reuse the primitive types from the original union schema, while for nested types, we want to use the `visitResult` because they have content from the previous recursive calls.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaVisitor.java
```diff
@@ -52,8 +52,21 @@
       case UNION:
         List<Schema> types = schema.getTypes();
         List<T> options = Lists.newArrayListWithExpectedSize(types.size());
-        for (Schema type : types) {
-          options.add(visit(type, visitor));
+        if (AvroSchemaUtil.isOptionSchema(schema)) {
```
shenodaguirguis, line 55: can you please verify that current definition of `isOptionSchema`, where it checks for null option in specific order/index, is still applicable here?

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -121,25 +125,76 @@ public void writeAndValidateOptionalComplexUnion() throws IOException {
         .project(expectedSchema)
         .build()) {
       rows = Lists.newArrayList(reader);
+
+      Assert.assertEquals("foo", rows.get(0).getStruct(0, 2).getString(1));
+      Assert.assertEquals(1, rows.get(1).getStruct(0, 2).getInt(0));
```
shenodaguirguis, line 138: can we verify it is `[foo, null]` and `[1, null]`  (here and everywhere; might worth a verification function)?

shenodaguirguis (approved):
LGTM, thanks for the fix @rzhang10 ! Left minor comments

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
rzhang10: it's actually `[null, foo]` and `[1, null]` and you can see I'm already validating this via `getString(1)` vs `getInt(0)` , see the difference index of 1 vs 0.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaVisitor.java
rzhang10: I'm not quite following this. I use `isOptionSchema` for the vanilla union case where it is composed of only 2 types one is null and one is the other type `[null, someType]` .

shenodaguirguis: I was concerned about `[someType, null]` but I recalled now that this gets reordered along the way, so we are all good.

core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
```diff
@@ -297,4 +298,19 @@ private static Schema copyRecord(Schema record, List<Schema.Field> newFields) {
   private static boolean isOptionSchemaWithNonNullFirstOption(Schema schema) {
     return AvroSchemaUtil.isOptionSchema(schema) && schema.getTypes().get(0).getType() != Schema.Type.NULL;
   }
+
+  // for primitive types, the visitResult will be null, we want to reuse the primitive types from the original
+  // schema, while for nested types, we want to use the visitResult because they have content from the previous
+  // recursive calls.
+  private static Schema copyUnion(Schema record, List<Schema> visitResults) {
```
funcheetah, line 305: What will ```null``` in ```[null, int, string]``` be in record and visitResults here?

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
shenodaguirguis: I am referring verifying that the other index 0/1 is `null`

shenodaguirguis: and may be also verify the size is 2?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -302,15 +306,28 @@ private UnionReader(List<ValueReader<?>> readers) {
...
+      }
+      InternalRow struct = new GenericInternalRow(nullIndex >= 0 ? alts.size() - 1 : alts.size());
+      for (int i = 0; i < struct.numFields(); i += 1) {
         struct.setNullAt(i);
       }
-      struct.update(index, value);
 
+      int index = decoder.readIndex();
+      Object value = this.readers[index].read(decoder, reuse);
+      if (nullIndex >= 0 && index > nullIndex) {
```
funcheetah: Could you please verify the case when `index == nullIndex`?

core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
rzhang10: It weill be both `null`.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
rzhang10: Added

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: I refactored the conditions to make it clearer, also added a null data record in `writeAndValidateOptionalComplexUnion` to test out.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -302,14 +306,31 @@ private UnionReader(List<ValueReader<?>> readers) {
...
+        Schema alt = alts.get(i);
+        if (Objects.equals(alt.getType(), Schema.Type.NULL)) {
+          nullIndex = i;
+          break;
+        }
+      }
+      InternalRow struct = new GenericInternalRow(nullIndex >= 0 ? alts.size() - 1 : alts.size());
+      for (int i = 0; i < struct.numFields(); i += 1) {
+        struct.setNullAt(i);
+      }
```
shenodaguirguis, line 322: for union `[string, null, int, long]` wouldn't that still lead to `[null, 1, null]` if, for example, the `int` option is set?

rzhang10: Yeah, but that is the expected struct right? what's the problem here

shenodaguirguis: > The previous implementation doesn't handle return values correctly when there is a null element in the union, like [null, int, string], it returned a struct data of [null, 1, null instead of just [1, null], this PR fixes it.

what is the desired behavior handling unions with null element?

rzhang10: To ignore the null element in the returned struct, so the returned struct has 1 size less than the count of the original union.

shenodaguirguis: gotcha! I think this should be clarified in this PR description.
Thanks @rzhang10 , this PR is all good to me.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -302,14 +306,31 @@ private UnionReader(List<ValueReader<?>> readers) {
...
       int index = decoder.readIndex();
       Object value = this.readers[index].read(decoder, reuse);
 
-      for (int i = 0; i < readers.length; i += 1) {
-        struct.setNullAt(i);
+      if (nullIndex < 0) {
+        struct.update(index, value);
+      } else if (index < nullIndex) {
+        struct.update(index, value);
+      } else if (index > nullIndex) {
```
funcheetah, line 331: Could you add comment on the case when ```index == nullIndex``` here? I think it is handled above but not obvious here given the branch.

rzhang10: for `index == nullIndex`, there should be no action, since the returned struct should be all null, and the previous code initialized the struct to be all null, so we can directly return it.

funcheetah: Yeah, I got the implementation. I'm saying that it is unclear when others look at this piece and may wonder how ```index == nullIndex``` is handled as the if condition skips this case. It is a nit, I'll leave it to you whether you want to add a comment here in code or not.

funcheetah (approved):
Nice PR. Thanks a lot @rzhang10 for fixing these issues.

PR #84, opened 2021-11-09
Avro: Change union read schema from hive to trino

rzhang10:
This rb changes the avro spark reader code path to read a union data type of `[int, string]` to a spark schema/data of `[tag, field0, field1]` instead of the previous `[tag_0, tag_1]`.
it also fixed an issue that when the avro union is a null, the reader should return a plain `null` instead of `[0, null, null]` or `[null, null]` (as the previous schema did).

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
+++ b/core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.avro;
 
+import java.util.ArrayList;
 import java.util.List;
 import org.apache.avro.LogicalType;
 import org.apache.avro.LogicalTypes;
@@ -116,12 +117,13 @@ class SchemaToType extends AvroSchemaVisitor<Type> {
       }
     } else {
       // Complex union
-      List<Types.NestedField> newFields = Lists.newArrayListWithExpectedSize(options.size());
+      List<Types.NestedField> newFields = new ArrayList<>();
+      newFields.add(Types.NestedField.required(allocateId(), "tag", Types.IntegerType.get()));
 
       int tagIndex = 0;
       for (Type type : options) {
         if (type != null) {
-          newFields.add(Types.NestedField.optional(allocateId(), "tag_" + tagIndex++, type));
+          newFields.add(Types.NestedField.optional(allocateId(), "field" + tagIndex++, type));
         }
       }
 
--- a/core/src/test/java/org/apache/iceberg/avro/TestUnionSchemaConversions.java
+++ b/core/src/test/java/org/apache/iceberg/avro/TestUnionSchemaConversions.java
@@ -25,7 +25,7 @@ import org.junit.Assert;
 import org.junit.Test;
 
 
-public class TestAvroComplexUnion {
+public class TestUnionSchemaConversions {
 
   @Test
   public void testRequiredComplexUnion() {
@@ -43,7 +43,8 @@ public class TestAvroComplexUnion {
 
     org.apache.iceberg.Schema icebergSchema = AvroSchemaUtil.toIceberg(avroSchema);
     String expectedIcebergSchema = "table {\n" +
-        "  0: unionCol: required struct<1: tag_0: optional int, 2: tag_1: optional string>\n" + "}";
+        "  0: unionCol: required struct<1: tag: required int, 2: field0: optional int, 3: field1: optional string>\n" +
+        "}";
 
     Assert.assertEquals(expectedIcebergSchema, icebergSchema.toString());
   }
@@ -65,32 +66,15 @@ public class TestAvroComplexUnion {
         .endRecord();
 
     org.apache.iceberg.Schema icebergSchema = AvroSchemaUtil.toIceberg(avroSchema);
-    String expectedIcebergSchema =
-        "table {\n" + "  0: unionCol: optional struct<1: tag_0: optional int, 2: tag_1: optional string>\n" + "}";
+    String expectedIcebergSchema = "table {\n" +
+        "  0: unionCol: optional struct<1: tag: required int, 2: field0: optional int, 3: field1: optional string>\n" +
+        "}";
 
     Assert.assertEquals(expectedIcebergSchema, icebergSchema.toString());
   }
 
   @Test
-  public void testSingleComponentUnion() {
-    Schema avroSchema = SchemaBuilder.record("root")
-        .fields()
-        .name("unionCol")
-        .type()
-        .unionOf()
-        .intType()
-        .endUnion()
-        .noDefault()
-        .endRecord();
-
-    org.apache.iceberg.Schema icebergSchema = AvroSchemaUtil.toIceberg(avroSchema);
-    String expectedIcebergSchema = "table {\n" + "  0: unionCol: required struct<1: tag_0: optional int>\n" + "}";
-
-    Assert.assertEquals(expectedIcebergSchema, icebergSchema.toString());
-  }
-
-  @Test
-  public void testOptionSchema() {
+  public void testSimpleUnionSchema() {
     Schema avroSchema = SchemaBuilder.record("root")
         .fields()
         .name("optionCol")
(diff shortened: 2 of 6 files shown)


funcheetah:
Thanks @rzhang10 for the PR! Could you please integrate test to make sure the new schema works?

core/src/test/java/org/apache/iceberg/avro/TestUnionSchemaConversions.java
```diff
@@ -108,22 +92,4 @@ public void testOptionSchema() {
 
     Assert.assertEquals(expectedIcebergSchema, icebergSchema.toString());
   }
-
```
autumnust, line 111: is this considered invalid at all? or what's the reason to get rid of this case ?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -292,7 +292,7 @@ protected void set(InternalRow struct, int pos, Object value) {
     }
   }
 
-  static class UnionReader implements ValueReader<InternalRow> {
+  private static class UnionReader implements ValueReader<Object> {
```
autumnust: is this modifier change necessary ? this might make it harder for us to contributing back in the future.

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
```diff
@@ -116,12 +117,13 @@ public Type union(Schema union, List<Type> options) {
       }
     } else {
       // Complex union
```
autumnust, line 119: is there an official definition of what is complex and what is simple ? If we target to contribute it back, maybe better to clarify that thru javadoc -- If this is merely internally usage I am fine with what we have now.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -316,20 +316,30 @@ public InternalRow read(Decoder decoder, Object reuse) throws IOException {
...
+      Object value = readers[index].read(decoder, reuse);
 
       if (nullIndex < 0) {
-        struct.update(index, value);
+        struct.update(index + 1, value);
+        struct.setInt(0, index);
       } else if (index < nullIndex) {
+        struct.update(index + 1, value);
+        struct.setInt(0, index);
+      } else {
```
autumnust, line 340: Not sure if I follow this, what does the relative position between value index and null index have anything to do with how we assign the value in the struct ?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
autumnust: do you also need to do the same for the vectorized orc reader?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -305,7 +305,7 @@ private UnionReader(Schema schema, List<ValueReader<?>> readers) {
     }
 
     @Override
-    public InternalRow read(Decoder decoder, Object reuse) throws IOException {
+    public Object read(Decoder decoder, Object reuse) throws IOException {
```
autumnust: why is this needed if you are still returning `struct` which is a `InternalRow` ?

autumnust (changes requested):

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: Because in avro the nullability of a union type is presented by the existence of a `NULL` type inside the union alternatives, and the `NULL` type can occur in any order inside the union, but in our converted struct, we don't have a field that corresponds to that `NULL` type, thus we have `-1` field count.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: See line 323, if it's null data, we will return `null` directly, so this method returns either `null` or an `InternalRow`, so the return type should only be `Object`. Please see the same method signature in `ValueReaders`.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: This mimics the same access control modifier for the similar method in this class `SparkValueReaders`, the relevant change has been made in `SparkOrcValueReaders` in pr #85 .

core/src/test/java/org/apache/iceberg/avro/TestUnionSchemaConversions.java
rzhang10: Yes, this schema itself shouldn't appear in the first place, user shouldn't define this kind of schema.

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
rzhang10: simple union is `[sometype, null]` while complex union is `[sometype1, sometype2, ...]`, where there are at least 2 non-null types in the union, I think we can probably add a java doc in `AvroSchemaUtil.isOptionSchema`, which is an existing method in upstream iceberg.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
autumnust: So why not returning null directly at line 323 if `nullIndex` is hit ?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
autumnust: I guess what I didn't know earlier was, if NULL type appears in the union it seems it has to be the first position in 
the alternative list ? But the avro spec said "Thus, for unions containing "null", the "null" is usually listed first," 

Will this lead to a silent failure if you don't check this first?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
autumnust: What about the modifier change ? Is that really needed?

autumnust:
A couple of more comments ... @rzhang10

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: `readers[index].read(decoder, reuse)` will return `null`, I just feel writing this way will make the code more consistent with later part.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: I think that's just a recommendation from the Avro spec, not mandated, in fact, in our ecosystem there are many avro schemas which failed to put `null` as the first type element in the union.

And that's why I'm specifically computing the null index here and branching on it.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: Yeah if you look at other nested class in this file you will see:
`private static class MapReader`
`private static class EnumReader`
`private static class ArrayReader`

so I think it's best to make the modifiers consistent.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
autumnust: Hmm not sure that's worthwhile with the cost of changing the method signature ... But I don't have strong opinion on this either.

autumnust (approved):
LGTM. Most of conservations above are just questions. I still believe we shouldn't change the signature of union read method from `InternalRow` to `Object` since a return type as `Object` doesn't seem to right if we can make it more specific. But I am no strong opinion on that and don't take that as blocker.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
rzhang10: I think using `InternalRow` is better, let me make the change accordingly. Thanks for spotting this!

PR #85, opened 2021-11-11
ORC: Change union read schema from hive to trino

rzhang10:
This rb changes the OC  spark reader  and vectorized reader code path to read a union data type of `[int, string]` to a spark schema/data of `[tag, field0, field1]` instead of the previous `[tag_0, tag_1]`.

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -311,9 +311,11 @@ public final class ORCSchemaUtil {
       Map<Integer, OrcField> mapping) {
     TypeDescription orcType;
     OrcField orcField = mapping.getOrDefault(fieldId, null);
+    // this branch means the iceberg struct schema actually correspond to an underlying union
     if (orcField != null && orcField.type.getCategory().equals(TypeDescription.Category.UNION)) {
       orcType = TypeDescription.createUnion();
-      for (Types.NestedField nestedField : type.asStructType().fields()) {
+      List<Types.NestedField> nestedFields = type.asStructType().fields();
+      for (Types.NestedField nestedField : nestedFields.subList(1, nestedFields.size())) {
         TypeDescription childType = buildOrcProjection(nestedField.fieldId(), nestedField.type(),
             isRequired && nestedField.isRequired(), mapping);
         orcType.addUnionChild(childType);
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
@@ -38,7 +38,7 @@ public abstract class OrcSchemaWithTypeVisitor<T> {
         return visitor.visitRecord(iType != null ? iType.asStructType() : null, schema, visitor);
 
       case UNION:
-        return visitUnion(iType, schema, visitor);
+        return visitor.visitUnion(iType, schema, visitor);
 
       case LIST:
         Types.ListType list = iType != null ? iType.asListType() : null;
@@ -71,12 +71,12 @@ public abstract class OrcSchemaWithTypeVisitor<T> {
     return visitor.record(struct, record, names, results);
   }
 
-  private static <T> T visitUnion(Type type, TypeDescription union, OrcSchemaWithTypeVisitor<T> visitor) {
+  protected T visitUnion(Type type, TypeDescription union, OrcSchemaWithTypeVisitor<T> visitor) {
     List<TypeDescription> types = union.getChildren();
     List<T> options = Lists.newArrayListWithCapacity(types.size());
 
     for (int i = 0; i < types.size(); i += 1) {
-      options.add(visit(type.asStructType().fields().get(i).type(), types.get(i), visitor));
+      options.add(visit(type.asStructType().fields().get(i + 1).type(), types.get(i), visitor));
     }
 
     return visitor.union(type, union, options);
--- a/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/VectorizedSparkOrcReaders.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/VectorizedSparkOrcReaders.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.spark.data.vectorized;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.IntStream;
@@ -36,6 +37,7 @@ import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
 import org.apache.orc.TypeDescription;
 import org.apache.orc.storage.ql.exec.vector.ListColumnVector;
+import org.apache.orc.storage.ql.exec.vector.LongColumnVector;
 import org.apache.orc.storage.ql.exec.vector.MapColumnVector;
 import org.apache.orc.storage.ql.exec.vector.StructColumnVector;
 import org.apache.orc.storage.ql.exec.vector.UnionColumnVector;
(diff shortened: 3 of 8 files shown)


orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
```diff
@@ -71,12 +71,12 @@ protected T visitRecord(
     return visitor.record(struct, record, names, results);
   }
 
-  private static <T> T visitUnion(Type type, TypeDescription union, OrcSchemaWithTypeVisitor<T> visitor) {
+  protected T visitUnion(Type type, TypeDescription union, OrcSchemaWithTypeVisitor<T> visitor) {
```
autumnust, line 74: why is this change needed?

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -311,9 +311,11 @@ private static TypeDescription buildOrcProjectForStructType(Integer fieldId, Typ
       Map<Integer, OrcField> mapping) {
     TypeDescription orcType;
     OrcField orcField = mapping.getOrDefault(fieldId, null);
+    // this branch means the iceberg struct schema actually correspond to an underlying union
     if (orcField != null && orcField.type.getCategory().equals(TypeDescription.Category.UNION)) {
       orcType = TypeDescription.createUnion();
-      for (Types.NestedField nestedField : type.asStructType().fields()) {
+      List<Types.NestedField> nestedFields = type.asStructType().fields();
```
autumnust, line 317: maybe precondition checking for the list size before accessing the second element and forwards?

spark/src/main/java/org/apache/iceberg/spark/data/vectorized/VectorizedSparkOrcReaders.java
```diff
@@ -443,11 +445,14 @@ public ColumnVector convert(org.apache.orc.storage.ql.exec.vector.ColumnVector v
         long batchOffsetInFile) {
       UnionColumnVector unionColumnVector = (UnionColumnVector) vector;
       List<Types.NestedField> fields = structType.fields();
-      assert fields.size() == unionColumnVector.fields.length;
```
autumnust, line 446: Shall we preserve this kind of assertion in terms of field size?

autumnust (changes requested):
Overall looks good but I still think we should gate empty union cases.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -311,9 +311,11 @@ private static TypeDescription buildOrcProjectForStructType(Integer fieldId, Typ
       Map<Integer, OrcField> mapping) {
     TypeDescription orcType;
     OrcField orcField = mapping.getOrDefault(fieldId, null);
+    // this branch means the iceberg struct schema actually correspond to an underlying union
```
funcheetah, line 314: nit: `corresponds`

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
```diff
@@ -115,106 +118,28 @@ public void testComplexUnion() throws IOException {
...
+      final Iterator<InternalRow> actualRowsIt = batchesToRows(reader.iterator());
 
-      Assert.assertEquals(columnarBatches.get(0).numRows(), NUM_OF_ROWS);
-      assertEquals(expectedSchema, expectedFirstRow, rowIterator.next());
-      assertEquals(expectedSchema, expectedSecondRow, rowIterator.next());
-    }
-  }
-
-  @Test
-  public void testSingleComponentUnion() throws IOException {
```
funcheetah, line 147: Why test for single component union is removed? How is `[ "string" ]` represented in the new schema?

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
funcheetah: +1 on why this change

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
```diff
@@ -225,23 +150,27 @@ public void testDeeplyNestedUnion() throws IOException {
 
     Schema expectedSchema = new Schema(
         Types.NestedField.optional(0, "c1", Types.StructType.of(
-            Types.NestedField.optional(1, "tag_0", Types.IntegerType.get()),
-            Types.NestedField.optional(2, "tag_1",
+            Types.NestedField.optional(100, "tag", Types.IntegerType.get()),
```
funcheetah, line 153: Could you explain how the value `100` is determined for the id field here?

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
rzhang10: This is just to align with the same code pattern of other visitor's method, see `visitRecord`, also, this change enables subclass overriding, which might be necessary in the future, since this `OrcSchemaWithTypeVisitor` is a very generic visitor.

spark/src/main/java/org/apache/iceberg/spark/data/vectorized/VectorizedSparkOrcReaders.java
rzhang10: I don't think it's necessary since I already know for sure the length would match, this is internal implementation detail.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
rzhang10: We already discussed before we won't support this schema and user should never use single type union cuz it doesn't make sense.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
rzhang10: it's just a random id I picked.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
funcheetah: Given it is a valid Avro schema, do we have a way to prevent user from creating this kinda of schema in the first place? What is the behavior if user do create it for current implementation?

rzhang10: The way to prevent it will be upon the due diligence of the user to not create such schema. We can just tell the user this case is not supported and behavior is undefined.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
rzhang10: Same as the comment below, I don't think it's necessary here since It's already known for sure the length would be at least 2, this is an internal implementation detail.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
rzhang10: Actually, I have a second thought on this, @autumnust when gobblin writes/converts an ORC dataset and encounters an Avor `[null, string]` type, does it write just `string` type to ORC file, or does it write a union? My intuition is that you guys write a single `string` as ORC schema is nullable by itself.

autumnust: So Gobblin does the transformation: https://github.com/apache/gobblin/blob/5bc22a5502ddeb810a35b0fb996dd9dfc8c81121/gobblin-utility/src/main/java/org/apache/gobblin/util/orc/AvroOrcSchemaConverter.java#L62 

So a default-value representation in Avro shall not leave in a state as single-element union.

funcheetah (approved):
LGTM. Regarding the single item union, I think we need to verify the behavior of it in current implementation. If we do not want to support this, we need a way to document/communicate that it is not supported.

PR #87, opened 2022-01-21
Separate classes under hive legacy package to new hivelink module

funcheetah:
This PR separates classes under hive legacy package to new hivelink module


build.gradle
```diff
@@ -495,6 +495,66 @@ project(':iceberg-hive-metastore') {
   }
 }
 
+project(':iceberg-hiveberg') {
```
wmoustafa: Are we modifying the root `build.gradle`? Why?

wmoustafa: Could you also clarify where these dependencies are copied from?

build.gradle
```diff
@@ -812,6 +872,7 @@ if (jdkVersion == '8') {
       compile project(':iceberg-arrow')
       compile project(':iceberg-hive-metastore')
       compile project(':iceberg-spark')
+      compile project(':iceberg-hiveberg')
```
wmoustafa: Is this an example of a circular dependency?

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalogs.java
```diff
@@ -35,20 +34,12 @@ private HiveCatalogs() {
...
 
-  private static final Cache<String, HiveCatalog> LEGACY_CATALOG_CACHE = Caffeine.newBuilder().build();
-
   public static HiveCatalog loadCatalog(Configuration conf) {
     // metastore URI can be null in local mode
     String metastoreUri = conf.get(HiveConf.ConfVars.METASTOREURIS.varname, "");
     return CATALOG_CACHE.get(metastoreUri, uri -> new HiveCatalog(conf));
   }
 
-  public static HiveCatalog loadLegacyCatalog(Configuration conf) {
```
wmoustafa, line 46: Is this a backward incompatible change? Do you have a plan on how to roll it out with Spark?

build.gradle
funcheetah: We are moving classes from legacy package to new module here as demo. This modification is needed to add a new module. If we move them to a new code repo, this change can be removed. All the dependencies are from hive-metastore `compileOnly` and `testCompile` dependencies, which are necessary and cannot be pulled by pulling hive-metastore module

build.gradle
funcheetah: This is not. This is needed since LegacyHiveTable used in `Reader` and `SparkBatchScan` is moved to `iceberg-hiveberg` module.

wmoustafa (commented):
Thanks @funcheetah for this PR. This is a great start.
Summary of offline discussion:
* Most of the remaining Iceberg changes are upgrading access modifiers (which is reasonable since we are extending the classes).
* Right now the dependency direction is from `iceberg-spark` to `hiveberg`. Assuming that direction is somewhat risky since the dependency should be the other way around. The only reason for having it this right now is that `Reader` and `SparkBatchScan` classes contain references to classes in `hiveberg` but there should be a way to overcome that by extending those classes and their calling chain (classes that instantiate them) up to a an SPI class, or converting them to SPI. @funcheetah to investigate that. `Reader` has been shown to be a simple example of extending since it called from the `IcebergSource` class, which implements a Spark SPI.
* We should develop a diff metric that reflects how close the code base is after each change to upstream Iceberg.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalogs.java
funcheetah: This is a backward incompatible change. It is not used directly from spark but from a wrapper layer. We can change the wrapper layer to accommodate this change.

hivelink/src/test/java/org/apache/iceberg/hivelink/ScriptRunner.java
```diff
@@ -0,0 +1,248 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iceberg.hivelink;
+
```
wmoustafa, line 21: This likely should be `com.linkedin.hivelink`.

wmoustafa (commented):
Given the package name change, could you please go through the deployment scenario and make sure deployment will be smooth for both Spark 2.3 and Spark 3.1? cc: @ljfgem since he considered similar backward compatibility issues in Coral.

hivelink/src/test/java/org/apache/iceberg/hivelink/ScriptRunner.java
funcheetah: I agree. As it is still a module in iceberg, we can keep it for now and change it to `com.linkedin.hivelink` once moved to a new repo.

funcheetah:
> Given the package name change, could you please go through the deployment scenario and make sure deployment will be smooth for both Spark 2.3 and Spark 3.1? cc: @ljfgem since he considered similar backward compatibility issues in Coral.

I think it should be fine as the dependencies change are captured on our wrapper layer.

wmoustafa (approved):

PR #89, opened 2022-02-01
Backport https://github.com/apache/iceberg/pull/2328 and its prerequisites 

autumnust:
## Identifying commits required 
To patch the PR of interests https://github.com/apache/iceberg/pull/2328 I identified four PRs that has a bunch of overlapping with it by scanning through the file git history of `HiveTableOperations.java` which is the file being modified the most. Specifically, the git history of this file in Li-Iceberg and Apache Iceberg are compared, from which I identified four PRs to be ported: 
https://github.com/apache/iceberg/pull/2123 [Push Iceberg table property values to HMS table properties]
https://github.com/apache/iceberg/pull/2252 [Change a key method's signature in  `HiveTableOperations.java` ]
https://github.com/apache/iceberg/pull/2263 [`acquireLock` method fixing in  `HiveTableOperations.java` ]
https://github.com/apache/iceberg/pull/2329 [Introduce total-files-size snapshot metric and populate HMS]

These are changes need to make `HiveTableOperations.java`  right. 
To logically replicate the fix into `HiveMetadataPreservingTableOperations.java`, did the following: 
- Change a couple of methods' access modifier to make them available in extended class, 
- Refactor HMS client code block into `persistTable` method, as well as the clean up method in finally block, 
- apply the try-catch of `checkCommitStatus` in the `doCommit` method, 
- remove `metadataUpdatedSuccessfully` which is no longer needed (and the original impl. is buggy since it only checks if the current version is exactly the same as new metadata version instead of checking if current version is in the lineage of new metadata version as the PR 2328 did). 

## Conflicts 
There's no conflicts if all four are ported, in the `HiveTableOperations.java` We have two commits that are not contributed back to upstream that: 
- simply remove the "private" access modifier from "unlock", "storageDescriptor", "acquireLock" and "setParameters" methods in the `HiveTableOperations.java`
- adding a code block for more log message in `HiveTableOperations.java` 

They are easily resolved. Overall, there should be minimal work to bring `HiveTableOperations.java` exactly the same as upstream.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveMetadataPreservingTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveMetadataPreservingTableOperations.java
@@ -36,9 +36,11 @@ import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.common.DynMethods;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.CommitFailedException;
+import org.apache.iceberg.exceptions.CommitStateUnknownException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.collect.ImmutableMap;
+import org.apache.iceberg.relocated.com.google.common.collect.ImmutableSet;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -46,6 +48,7 @@ import org.slf4j.LoggerFactory;
 
 /**
  * A {@link HiveTableOperations} that does not override any existing Hive metadata.
+ * TODO: This extension should be removed once dual-publish of iceberg+hive is stopped.
  *
  * The behaviour of this class differs from {@link HiveTableOperations} in the following ways:
  * 1. Does not modify serde information of existing Hive table, this means that if Iceberg schema is updated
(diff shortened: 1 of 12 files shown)


shenodaguirguis:
thanks @autumnust  for the quick rb, and the documentation of the changes!
Checks are failing with dependency issue. You didn't change dependencies though, but I re-ran the checks and it still failed. Can you please take a look? Otherwise, it looks good to me.

rzhang10:
> thanks @autumnust for the quick rb, and the documentation of the changes! Checks are failing with dependency issue. You didn't change dependencies though, but I re-ran the checks and it still failed. Can you please take a look? Otherwise, it looks good to me.

Seems like the error has to do with "palantir.bintray.com" and I see https://github.com/linkedin/iceberg/pull/88/files removed them, maybe rebase on that and see if it works?

autumnust:
Rebased, and logically applied the fix in  `HiveMetadataPreservingTableOperations.java`
Can you take a second look ? @shenodaguirguis  @rzhang10  

FYI @ZihanLi58  @jack-moseley

autumnust:
Also, I decided to not to take down the patch in the HiveTableOperations since they are needed anyway once we decide to move to Apache Iceberg I assume.  This also reduces the amount of code that I need to duplicate.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveMetadataPreservingTableOperations.java
```diff
@@ -180,26 +183,27 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+        LOG.error("Cannot tell if commit to {}.{} succeeded, attempting to reconnect and check.",
+            database, tableName, persistFailure);
+        commitStatus = checkCommitStatus(newMetadataLocation, metadata);
+        switch (commitStatus) {
+          case SUCCESS:
+            break;
+          case FAILURE:
+            throw persistFailure;
+          case UNKNOWN:
+            throw new CommitStateUnknownException(persistFailure);
```
shenodaguirguis, line 204: does it make sense to log here too, given the log at line 195? 
If you agree, I think that would apply to `hiveTableOperations.java` too

shenodaguirguis (approved):
thanks @autumnust ! Left a minor comment, otherwise LGTM!

rzhang10:
Overall LGTM.

ZihanLi58 (approved):
+1, thanks for helping fix this!

PR #90, opened 2022-02-04
Call unlock lockId before throwing an exception in acquireLock method

maluchari:
In the [cleans up](https://github.com/linkedin/iceberg/blob/2c59857a4f8bde052121d2edb562ed83d135863f/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L381.)  path  lockId is not returned in [code](https://github.com/linkedin/iceberg/blob/2c59857a4f8bde052121d2edb562ed83d135863f/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L368) for cleanup and hence becomes a no-op during unlock.

### DIFF
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -357,11 +357,15 @@ public class HiveTableOperations extends BaseMetastoreTableOperations {
 
     // timeout and do not have lock acquired
     if (timeout && !state.get().equals(LockState.ACQUIRED)) {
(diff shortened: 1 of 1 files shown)


rzhang10:
I have several points here:

- these 2 branches are when the lock state is waiting or is not required, this means other processes/threads must have hold the lock. So it's actually inaccurate if we try to release the locks on their behalf?
- it seems our hive doesn't enable concurrency currently, so seems this entire locking/unlocking on HMS might just have no effect to begin with. I doubt the original issue is really caused by this.

maluchari:
Closing this as this seems to have been fixed upstream in [this PR](https://github.com/apache/iceberg/pull/2263)

PR #91, opened 2022-02-17
Support timestamp as partition type

ljfgem:
This PR enhances https://github.com/linkedin/iceberg/pull/78 to backport the [upstream PR](https://github.com/apache/iceberg/pull/3933) to support timestamp as a partition type. 

Hard to add unit tests given we don't have corresponding `TestAddFilesProcedure` test class in li-0.11.x branch.
Tested on our Spark shell and it worked well.

### DIFF
--- a/api/src/main/java/org/apache/iceberg/types/Conversions.java
+++ b/api/src/main/java/org/apache/iceberg/types/Conversions.java
@@ -28,8 +28,11 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.StandardCharsets;
+import java.sql.Timestamp;
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import org.apache.iceberg.exceptions.RuntimeIOException;
 import org.apache.iceberg.expressions.Literal;
 import org.apache.iceberg.util.UUIDUtil;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/apache/iceberg/types/Conversions.java
```diff
@@ -72,7 +75,12 @@ public static Object fromPartitionString(Type type, String asString) {
       case DATE:
         return Literal.of(asString).to(Types.DateType.get()).value();
       case TIMESTAMP:
-        return Literal.of(asString).to(Types.TimestampType.withoutZone()).value();
+        if (!asString.contains("T")) {
+          final Instant instant = Timestamp.valueOf(asString).toInstant();
+          return TimeUnit.SECONDS.toMicros(instant.getEpochSecond()) + TimeUnit.NANOSECONDS.toMicros(instant.getNano());
```
rzhang10: How do you think we use `asString.replaceFirst(" ", "T")` to make it a ISO standard string that can be parsed?

ljfgem: I think both can work, and the current one has been verified, could you share the advantage of doing `asString.replaceFirst(" ", "T")`?

rzhang10: It's just simpler and you don't need to have those imports.

ljfgem: Thanks, modified and tested

funcheetah (approved):
Thanks @ljfgem . LGTM.

PR #92, opened 2022-03-02
[LI] Align default value validation align with avro semantics in terms of nullable (nested) fields

rzhang10:
Avro record level default values behave this way:

say we have a avro schema:
```
record: {
  f1: string,
  f2: int
}
```
its default value should defined sth like:
```
{f1: "foo", f2: 1}
```

(Case-1)
But when the inner field `f1` is nullable, i.e. optional with the default value null, then the record level default value can be defined as below, omitting the default value for it.
```
{f2: 1}
```

(Case-2)
Since the inner field `f1` is nullable, its default value should be `null` as always, however, it's also fine to repeat this default value in the record default, as:
```
{f1: null, f2: 1}
```

This change is to cover these 2 cases and allow the schema validation for default value to pass.

### DIFF
--- a/core/src/test/java/org/apache/iceberg/avro/TestSchemaConversions.java
+++ b/core/src/test/java/org/apache/iceberg/avro/TestSchemaConversions.java
@@ -325,4 +325,52 @@ public class TestSchemaConversions {
     List<String> origFieldDocs = Lists.newArrayList(Iterables.transform(origSchema.columns(), Types.NestedField::doc));
     Assert.assertEquals(origFieldDocs, fieldDocs);
   }
+
+  @Test
+  public void testConversionOfRecordDefaultWithOptionalNestedField() {
+    String schemaString = "{\n" +
+            "    \"type\": \"record\",\n" +
+            "    \"name\": \"root\",\n" +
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/iceberg/avro/TestSchemaConversions.java
```diff
@@ -325,4 +325,52 @@ public void testFieldDocsArePreserved() {
...
+            "                ]\n" +
+            "            },\n" +
+            "            \"default\": {\n" +
+            "                \"mapField\": {}\n" +
+            "            }\n" +
+            "        }\n" +
+            "    ]\n" +
+            "}";
+    Schema schema = new Schema.Parser().parse(schemaString);
+    AvroSchemaUtil.toIceberg(schema);
```
funcheetah: Could you add assertion to unit test?

rzhang10: Added.

funcheetah (approved):
Thanks @rzhang10 for the fix. LGTM

PR #93, opened 2022-03-16
[LI][Avro] Do not reorder elements inside a Avro union schema

rzhang10:
This is basically to revert upstream https://github.com/apache/iceberg/pull/1132 . The upstream PR fixed the issue by explicitly setting all Avro default to null, since in upstream default value in Avro doesn't affect any Iceberg behavior, so removing them would not cause any effect and it make the schema to always align with the first null type in union schema. The "making null first type in union" sanitizing logic is in  `PruneColumns.union`.

However, after we add avro-based default value feature in LI-Iceberg, an optional field can also have a default value in Iceberg, and the avro default shouldn't be discarded (setting to null).

This means an avro schema of `["int", "null"]` with default value `1` is itself of significance and should be kept as is, this PR is to make that happen.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
+++ b/core/src/main/java/org/apache/iceberg/avro/PruneColumns.java
@@ -22,7 +22,6 @@ package org.apache.iceberg.avro;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import org.apache.avro.JsonProperties;
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaNormalization;
 import org.apache.iceberg.mapping.NameMapping;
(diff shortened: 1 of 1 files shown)


wmoustafa:
A few questions:
1- The upstream PR had a test file. Should that be reverted too?
2- I understand the upstream PR will effectively convert the default to NULL. Is my understanding correct that this PR will preserve the default value instead?
3- How was default value preserved earlier if this would always come in the way?
4- I think we should add a test case to accompany that code change.

If it helps to update the description with the answers to some of those questions, please do.

rzhang10:
> A few questions: 1- The upstream PR had a test file. Should that be reverted too? 2- I understand the upstream PR will effectively convert the default to NULL. Is my understanding correct that this PR will preserve the default value instead? 3- How was default value preserved earlier if this would always come in the way? 4- I think we should add a test case to accompany that code change.
> 
> If it helps to update the description with the answers to some of those questions, please do.

1. It does not need to because we also removed the reordering logic, the test can still pass because the avro schema will still be valid now that the first type in union matches the default value's type
2. Yes
3. Without this patch, reading tables of such schema in production will throw exception
4. I think adding a test might not be necessary, because we eventually will switch to apache/iceberg which we can get default value feature for free, and this PR that manipulate the avro schema (and any tests) can be discarded anyways, because the future upstream default value is preserved in the Iceberg schema itself, and is completely independent of Avro schema.

wmoustafa (approved):

rzhang10:
Also just to reassure, I tested this change on gateway and it worked for the problematic tables.

PR #94, opened 2022-03-16
[LI][Spark][Avro] read avro union using decoder instead of directly returning v…

rzhang10:
…alue

Previously the code directly will directly return `null`, although the value it returns is correct, it ignores to make the correct internal Avro decoder/parser state transitions. Using the `reader.read(decoder, reuse)` is the correct way to transition the internal Avro parser state. I've also created a nested schema test case that will fail without this fix.

### DIFF
--- a/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
+++ b/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
@@ -279,6 +279,57 @@ public class TestSparkAvroUnions {
 
   @Test
   public void testDeeplyNestedUnionSchema3() throws IOException {
+    /*
+    * the printed write schema:
+    * {
+      "type": "record",
+      "name": "root",
+      "fields": [
+        {
+          "name": "value",
+          "type": [
+            {
+              "type": "record",
+              "name": "r1",
+              "fields": [
+                {
+                  "name": "ff1",
+                  "type": "long"
+                },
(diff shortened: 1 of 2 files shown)


spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -275,4 +276,92 @@ public void testDeeplyNestedUnionSchema2() throws IOException {
       Assert.assertEquals(1, rows.get(0).getArray(0).getStruct(0, 3).getStruct(1, 1).getInt(0));
     }
   }
+
+  @Test
+  public void testDeeplyNestedUnionSchema3() throws IOException {
+    org.apache.avro.Schema writeSchema = SchemaBuilder
```
funcheetah, line 333: Do you think it makes sense to add a comment for the string format schema? It is hard to read the actual schema this test is testing against.

funcheetah (commented):
Thanks @rzhang10 for the patch.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
rzhang10: Sure, I will add a pretty printed json string version of the schema as a comment.

funcheetah (approved):
Thanks @rzhang10 ! LGTM

PR #96, opened 2022-04-13
Improve the logging when the deserailzed index is invalid to read the symbol from enum

yiqiangin:
**Problem**
According to a ticket issued by a user,  `ArrayIndexOutOfBoundsException` is thrown from the code of the class of `SparkValueReaders$EnumReader` when a Spark job is executed. It happens when an enum evolves with new symbols, EnumReader tries to use the index of the new symbol to read the enum before the evolution which results in an `ArrayIndexOutOfBoundsException`.

**Solution**
In EnumReader.read(), it checks if the value of the index deserialized for a symbol is in the bound of the number of the symbols in the target enum. If not, write an error log to help the investigation by the users and return null. If yes, read the symbol from enum.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -998,10 +998,6 @@ project(':iceberg-spark3') {
     testCompile project(path: ':iceberg-hive-metastore', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-api', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-data', configuration: 'testArtifacts')
-    //snappy-java supports Apple M1 cpu at the version of 1.1.8.2 when the tests are executed with JDK 11
-    if(JavaVersion.current() == JavaVersion.VERSION_11){
-      testRuntime "org.xerial.snappy:snappy-java:1.1.8.2"
-    }
   }
 
   test {
@@ -1048,10 +1044,6 @@ project(":iceberg-spark3-extensions") {
     testCompile project(path: ':iceberg-hive-metastore', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-spark', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-spark3', configuration: 'testArtifacts')
-    //snappy-java supports Apple M1 cpu at the version of 1.1.8.2 when the tests are executed with JDK 11
-    if(JavaVersion.current() == JavaVersion.VERSION_11){
-      testRuntime "org.xerial.snappy:snappy-java:1.1.8.2"
-    }
 
     // Required because we remove antlr plugin dependencies from the compile configuration, see note above
     runtime "org.antlr:antlr4-runtime:4.7.1"
--- a/build.gradle
+++ b/build.gradle
@@ -998,6 +998,10 @@ project(':iceberg-spark3') {
     testCompile project(path: ':iceberg-hive-metastore', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-api', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-data', configuration: 'testArtifacts')
+    //snappy-java supports Apple M1 cpu at the version of 1.1.8.2 when the tests are executed with JDK 11
+    if(JavaVersion.current() == JavaVersion.VERSION_11){
+      testRuntime "org.xerial.snappy:snappy-java:1.1.8.2"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -998,6 +998,10 @@ project(':iceberg-spark3') {
     testCompile project(path: ':iceberg-hive-metastore', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-api', configuration: 'testArtifacts')
     testCompile project(path: ':iceberg-data', configuration: 'testArtifacts')
+    //snappy-java supports Apple M1 cpu at the version of 1.1.8.2 when the tests are executed with JDK 11
```
wmoustafa: Does it run successfully with Java 8? Maybe Java 11 is not supported yet?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -123,7 +127,12 @@ private EnumReader(List<String> symbols) {
     @Override
     public UTF8String read(Decoder decoder, Object ignore) throws IOException {
       int index = decoder.readEnum();
-      return symbols[index];
+      if (index < 0 || index >= symbols.length) {
+        LOG.error("Unable to read the symbol in the given enum as the deserialized index {} is out of bound", index);
+        return null;
```
wmoustafa: Should not we throw an exception for backward compatibility? Also, `null` could cause silent data errors?

wmoustafa: It is not clear to me if index could ever be less than zero, so we may skip that check.

build.gradle
yiqiangin: Yes, it can be executed successfully with Java 8 in IntelliJ. Do you think this fix is still necessary in build.gradle?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
wmoustafa: My concern is that it may not be accepted upstream if Java 11 is indeed not supported. I would skip it and ask to use Java 8. You can also try to find the answer to which Java is supported in the documentation to see what the official support is.

build.gradle
wmoustafa: My concern is that it may not be accepted upstream if Java 11 is indeed not supported. I would skip it and ask to use Java 8. You can also try to find the answer to which Java is supported in the documentation to see what the official support is.

yiqiangin: It supports Java 11 as shown in https://github.com/apache/iceberg. I have also tried to build different branches from Apache repo with Java 11. It succeeds in the branches of 0.12.x, 0.13.x and master, as the version of `snappy-java` has already been upgraded to 1.1.8.2 according to the result of checking dependencies with gradle command.
As our branch of li-0.11.x is forked from the branch of 0.11.x, I also tried to build this branch from Apache repo, however it fails because the dependency of guava-28.0-jre cannot be resolved. I guess the branch of 0.11.x in Apache repo has not been maintained for a long time.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
wmoustafa: Thanks for looking into this! Seems the problem is that we are on an old version of Iceberg. To minimize diversion with Apache Iceberg, I would keep it and get it for free on the upgrade. One of the programs we are running now in the group is to stick to Apache Iceberg and not diverge as much as possible.

build.gradle
wmoustafa: Thanks for looking into this! Seems the problem is that we are on an old version of Iceberg. To minimize diversion with Apache Iceberg, I would keep it and get it for free on the upgrade. One of the programs we are running now in the group is to stick to Apache Iceberg and not diverge as much as possible.

yiqiangin: Sure, this change is removed.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
yiqiangin: Done.

wmoustafa (approved):

PR #98, opened 2022-04-20
Handle non-nullable union of single type for Avro

funcheetah:
This PR handles non-nullable union of single type to be consistent with Trino representation for Avro.

Non-nullable single type union ```[type]``` should be read as ```type``` instead of ```struct<tag, field0>```

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
@@ -139,9 +139,7 @@ public class AvroSchemaUtil {
   }
 
   /**
-   * This method decides whether a schema is single type union
-   *
-   * single type union: a schema is of type union and there is only one option
+   * This method decides whether a schema represents a single type union, i.e., a union that contains only one option
    *
    * @param schema input schema
    * @return true if schema is single type union
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
@@ -139,6 +139,22 @@ public class AvroSchemaUtil {
   }
 
   /**
+   * This method decides whether a schema is single type union
+   *
+   * single type union: a schema is of type union and there is only one option
+   *
+   * @param schema input schema
+   * @return true if schema is single type union
+   */
+  public static boolean isSingleTypeUnion(Schema schema) {
+    if (schema.getType() == UNION && schema.getTypes().size() == 1) {
+      return true;
+    }
+
+    return false;
+  }
+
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
```diff
@@ -138,6 +138,22 @@ public static boolean isOptionSchema(Schema schema) {
...
+   *
+   * @param schema input schema
+   * @return true if schema is single type union
+   */
+  public static boolean isSingleTypeUnion(Schema schema) {
+    if (schema.getType() == UNION && schema.getTypes().size() == 1) {
+      return true;
+    }
+
+    return false;
```
xkrogen: We can simplify to just:
```suggestion
    return schema.getType() == UNION && schema.getTypes().size() == 1;
```

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
```diff
@@ -141,6 +141,44 @@ public void writeAndValidateOptionalComplexUnion() throws IOException {
 
   @Test
   public void writeAndValidateSingleTypeUnion() throws IOException {
```
rzhang10, line 143: Could you add another test where the avro schema is a single union of a complex type (e.g. struct/map/array) ?

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
funcheetah: Thanks for the suggestions. Updated the method.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroUnions.java
funcheetah: Sure, will add test case for it.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
```diff
@@ -138,6 +138,18 @@ public static boolean isOptionSchema(Schema schema) {
     return false;
   }
 
+  /**
+   * This method decides whether a schema is single type union
+   *
+   * single type union: a schema is of type union and there is only one option
```
wmoustafa: `This method decides whether a schema represents a single type union, i.e., a union that contains only one option` (and we can remove the second line).

funcheetah: Thanks for the suggestion. Updated.

wmoustafa (approved):

PR #102, opened 2022-05-05
Move 'Hive Metadata Scan: Support case insensitive name mapping' (PR 52) to hivelink-core

ljfgem:
This PR is to move [Hive Metadata Scan: Support case insensitive name mapping](https://github.com/linkedin/iceberg/commit/78a1208b3c8311a32b9e5ec74e413f47264eed0b) to hivelink-core.
Given the changes in [NameMapping](https://github.com/linkedin/iceberg/commit/78a1208b3c8311a32b9e5ec74e413f47264eed0b#diff-d0bbb52afd4e54f3e77d4e93c36b501c66617bd2b07a48e8f10228d324c7226bR48) and [NameMappingParser](https://github.com/linkedin/iceberg/commit/78a1208b3c8311a32b9e5ec74e413f47264eed0b#diff-4a57712294724f3685298e645c209565af4a0148a560b583eb20766c34312779R80) are tightly coupled, this PR doesn't move the changes from these 2 files.

Tested by building it.

### DIFF
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTableOperations.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTableOperations.java
@@ -71,9 +71,9 @@ import org.slf4j.LoggerFactory;
 public class LegacyHiveTableOperations extends BaseMetastoreTableOperations {
 
   private static final Logger LOG = LoggerFactory.getLogger(LegacyHiveTableOperations.class);
-  static final long INITIAL_SEQUENCE_NUMBER = 0;
-  static final int DEFAULT_TABLE_FORMAT_VERSION = 1;
-  static final int INITIAL_SPEC_ID = 0;
+  private static final long INITIAL_SEQUENCE_NUMBER = 0;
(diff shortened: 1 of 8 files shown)


hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTableOperations.java
```diff
@@ -68,6 +71,9 @@
 public class LegacyHiveTableOperations extends BaseMetastoreTableOperations {
 
   private static final Logger LOG = LoggerFactory.getLogger(LegacyHiveTableOperations.class);
+  static final long INITIAL_SEQUENCE_NUMBER = 0;
```
yiqiangin: why not declare it as `private` for these three constant variables?

yiqiangin (approved):
only one minor comment.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTableOperations.java
ljfgem: Thanks for pointing it out, modified them.

PR #105, opened 2022-05-11
Modify support for reading timestamp without timezone to be aligned with upstream

ljfgem:
This PR is to modify [this commit](https://github.com/linkedin/iceberg/commit/c7f97aefff3d201cd949b8a55f833eec97346d8c) to be aligned with the [upstream PR](https://github.com/apache/iceberg/pull/2757/files), mainly in the following aspects:
1. Add support for writing timestamp without timezone
2. Move previous `READ_TIMESTAMP_WITHOUT_ZONE` flag to `HANDLE_TIMESTAMP_WITHOUT_TIMEZONE` in `SparkUtil`. We still keep using `"read-timestamp-without-zone"` as the value of `HANDLE_TIMESTAMP_WITHOUT_TIMEZONE` so that we don't need to modify its value internally
3. Add more tests

Tests:
1. `./gradlew clean build`
2. All the existing and new unit tests
3. Tested on gateway, worked well

### DIFF
--- a/spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java
@@ -127,6 +127,30 @@ public class SparkSchemaUtil {
   }
 
   /**
+   * Convert a Spark {@link StructType struct} to a {@link Schema} with new field ids.
+   * <p>
+   * This conversion assigns fresh ids.
+   * <p>
+   * Some data types are represented as the same Spark type. These are converted to a default type.
+   * <p>
+   * To convert using a reference schema for field ids and ambiguous types, use
+   * {@link #convert(Schema, StructType)}.
+   *
+   * @param sparkType a Spark StructType
+   * @param useTimestampWithoutZone boolean flag indicates that timestamp should be stored without timezone
+   * @return the equivalent Schema
+   * @throws IllegalArgumentException if the type cannot be converted
+   */
+  public static Schema convert(StructType sparkType, boolean useTimestampWithoutZone) {
+    Type converted = SparkTypeVisitor.visit(sparkType, new SparkTypeToType(sparkType));
+    Schema schema = new Schema(converted.asNestedType().asStructType().fields());
+    if (useTimestampWithoutZone) {
+      schema = SparkFixupTimestampType.fixup(schema);
+    }
+    return schema;
+  }
+
+  /**
    * Convert a Spark {@link DataType struct} to a {@link Type} with new field ids.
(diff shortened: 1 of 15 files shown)


rzhang10:
Why do we need this patch, wouldn't it be when we switch to apache/iceberg we will get the upstream PR for free?

If `Add support for writing timestamp without timezone` is a new feature in addition to the upstream PR, maybe we can contribute it to upstream?

ljfgem:
> Why do we need this patch, wouldn't it be when we switch to apache/iceberg we will get the upstream PR for free?
> 
> If `Add support for writing timestamp without timezone` is a new feature in addition to the upstream PR, maybe we can contribute it to upstream?

This PR was only committed to upstream master branch and upstream 0.11.x branch doesn't have this PR, and I don't think we are able to migrate to upstream master branch soon. So this PR is mainly to reduce the diff between li-0.11.x and upstream master branch.

spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java
```diff
@@ -126,6 +126,30 @@ public static Schema convert(StructType sparkType) {
...
+   *
+   * @param sparkType a Spark StructType
+   * @param useTimestampWithoutZone boolean flag indicates that timestamp should be stored without timezone
+   * @return the equivalent Schema
+   * @throws IllegalArgumentException if the type cannot be converted
+   */
+  public static Schema convert(StructType sparkType, boolean useTimestampWithoutZone) {
+    Type converted = SparkTypeVisitor.visit(sparkType, new SparkTypeToType(sparkType));
+    Schema schema = new Schema(converted.asNestedType().asStructType().fields());
+    if (useTimestampWithoutZone) {
```
yiqiangin, line 147: minor: using `return useTimestampWithoutZone ? SparkFixupTimestampType.fixup(schema) : schema` instead to make it more concise.

spark/src/test/java/org/apache/iceberg/spark/data/parquet/vectorized/TestParquetVectorizedReads.java
```diff
@@ -260,4 +260,9 @@ public void testReadsForTypePromotedColumns() throws Exception {
     assertRecordsMatch(readSchema, 30000, data, dataFile, false,
         true, BATCH_SIZE);
   }
+
```
yiqiangin, line 263: Do you miss the implementation of this test?

yiqiangin (approved):
Codewise it looks good to me. Pls make sure this PR is necessary before it is merged.

PR #106, opened 2022-05-12
Disable default value preserving

rzhang10:
This patch disable the LI-Iceberg implementation of default value preserving.
Since the related code exposes several bugs and impact production, and since in our current environment we don't really rely on Iceberg metadata. So I feel disabling this piece is the best short term solution.

In long term, after we switch to apache/iceberg and start using iceberg metadata, we can get default value feature for free https://github.com/apache/iceberg/pull/4301 so the implementation in LI-Iceberg is irrelevant at that time.


funcheetah (approved):

rzhang10:
Seems the CI encountered a transient issue we had before:
`org.apache.thrift.transport.TTransportException: java.net.SocketException: Broken pipe (Write failed)`
and only the java 11 run failed, so I think it can safely merge now.

PR #107, opened 2022-05-13
Handle single type union for deep nested schema

funcheetah:
This PR fixes an issue for deep nested single type union in https://github.com/linkedin/iceberg/pull/98


rzhang10:
@funcheetah thanks for fixing it, LGTM!

PR #108, opened 2022-05-16
Column projection of union type

yiqiangin:
**Problem**
Currently column projection does not work for union type. The root cause of the problem is as follows: 
- The current code assumes that the types inside a union from Avro schema should match the type fields inside a union struct from Iceberg schema when Avro union reader is created.
-  In case of column projection of union type, the current code only prune the schema of union in Iceberg schema with the projected fields, while the union of Avro schema still contains all the types. It results in the mismatch between Avro schema and Iceberg schema for the union in this case. 
- However, as all the contents of each data type in a union in Avro file should be read by Avro readers correctly no matter this data type is projected or not based on the decoding procedure of Avro file, all the types in a union from Avro schema are needed to create the corresponding type readers in AvroUnionReader even in case of column projection. Therefore the union in Avro schema cannot be pruned like what is done to the union struct in Iceberg schema.

**Solution**
In addition, the order of the types in a union in Iceberg schema can be identified from its field name like "field0".."fieldK". K is the index which can be used to match the order of the types in the union of Avro schema.

In the code of create the readers of all types in the union of Avro schema (namely `AvroSchemaWithTypeVisitor.visitUnion`), checking the fields of the struct corresponding union in Iceberg schema to create a map between the order index and the field type in Iceberg schema. When iterating through all the types in Avro schema, using the order index to check if the corresponding type exists in the map, if yes which means the field is projected, creating the option of creating the reader with the type in Iceberg schema, otherwise, creating the option with type null. 

In the code of AvroUnionReader, Iceberg schema needs to be passed into it. The fields of the returned row should be constructed based on the fields in Iceberg schema not the types in Avro schema. If tag field is projected, one more field is added in the beginning of the row and updated with the index of the field in Avro file.

**Test**
All the test cases in `TestSparkAvroUnions.java` with a new test case `writeAndValidateRequiredComplexUnionWithProjection`

```
val df = spark.sql("select c1.field0 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.field0,c1.field1 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.tag,c1.field0 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.tag,c1.field0,c1.field1 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.field1,c1.field0,c1.tag from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.field1,c1.field0 from u_yiqding.avro_union_table_test")
```

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/Avro.java
+++ b/core/src/main/java/org/apache/iceberg/avro/Avro.java
@@ -385,6 +385,7 @@ public class Avro {
     };
     private Long start = null;
     private Long length = null;
+    private Schema fileSchema = null;
 
     private ReadBuilder(InputFile file) {
       Preconditions.checkNotNull(file, "Input file cannot be null");
@@ -446,6 +447,11 @@ public class Avro {
       return this;
     }
 
+    public ReadBuilder setFileSchema(Schema fileSchema) {
+      this.fileSchema = fileSchema;
+      return this;
+    }
+
     public <D> AvroIterable<D> build() {
       Preconditions.checkNotNull(schema, "Schema is required");
       Function<Schema, DatumReader<?>> readerFunc;
@@ -458,7 +464,7 @@ public class Avro {
       }
 
       return new AvroIterable<>(file,
-          new ProjectionDatumReader<>(readerFunc, schema, renames, nameMapping),
+          new ProjectionDatumReader<>(readerFunc, schema, renames, nameMapping, fileSchema),
           start, length, reuseContainers);
     }
   }
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -106,11 +106,13 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
 
   /*
   A complex union with multiple types of Avro schema is converted into a struct with multiple fields of Iceberg schema.
-  A field is related to a type in the order defined in Avro schema. Also, an extra tag field is added into the struct of
-  Iceberg schema. The user can query the column of union type with the field projected (e.g. colUnion.field0) in which
-  the maximum number of the fields to be projected equals to the number of fields of the complete struct converted
-  from the union. The case of without field projection equals to the case of full fields projection.
-  Therefore, this function visits the complex union by assuming the field projection always happens.
+  Also an extra tag field is added into the struct of Iceberg schema during the conversion.
+  The fields in the struct of Iceberg schema are expected to be stored in the same order
+  as the corresponding types in the union of Avro schema.
+  Except the tag field, the fields in the struct of Iceberg schema are the same as the types in the union of Avro schema
+  in the general case. In case of field projection, the fields in the struct of Iceberg schema only contains
+  the fields to be projected which equals to a subset of the types in the union of Avro schema.
+  Therefore, this function visits the complex union with the consideration of both cases.
    */
   private static <T> void visitComplexUnion(Type type, Schema union,
                                             AvroSchemaWithTypeVisitor<T> visitor, List<T> options) {
@@ -124,37 +126,35 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
       if (schema.getType() == Schema.Type.NULL) {
         nullTypeFound = true;
         options.add(visit((Type) null, schema, visitor));
-        typeIndex++;
-        continue;
-      }
-
-      // If a NULL type is found before current type, the type index is one larger than the actual type index which
-      // can be used to track the corresponding field in the struct of Iceberg schema.
-      int actualTypeIndex = nullTypeFound ? typeIndex - 1 : typeIndex;
-      boolean relatedFieldInStructFound = false;
-      while (fieldIndexInStruct < type.asStructType().fields().size()) {
-        String structFieldName = type.asStructType().fields().get(fieldIndexInStruct).name();
-        if (UNION_TAG_FIELD_NAME.equals(structFieldName)) {
-          fieldIndexInStruct++;
-          continue;
-        }
-
-        int indexFromStructFieldName = Integer.valueOf(structFieldName.substring(5));
-        if (actualTypeIndex == indexFromStructFieldName) {
-          relatedFieldInStructFound = true;
-          options.add(visit(type.asStructType().fields().get(fieldIndexInStruct).type(), schema, visitor));
+      } else {
+        boolean relatedFieldInStructFound = false;
+        Types.StructType struct = type.asStructType();
+        if (fieldIndexInStruct < struct.fields().size() &&
+            UNION_TAG_FIELD_NAME.equals(struct.fields().get(fieldIndexInStruct).name())) {
           fieldIndexInStruct++;
         }
-        break;
-      }
 
-      // If a field is not projected, a corresponding field in the struct of Iceberg schema cannot be found
-      // for current type of union in Avro schema, a reader for current type still needs to be created and
-      // used to make the reading of Avro file successfully. In this case, a null field type is used to
-      // create the option for the reader of the current type which still can read the corresponding content
-      // in Avro file successfully.
-      if (!relatedFieldInStructFound) {
-        options.add(visit((Type) null, schema, visitor));
+        if (fieldIndexInStruct < struct.fields().size()) {
+          // If a NULL type is found before current type, the type index is one larger than the actual type index which
+          // can be used to track the corresponding field in the struct of Iceberg schema.
+          int actualTypeIndex = nullTypeFound ? typeIndex - 1 : typeIndex;
+          String structFieldName = type.asStructType().fields().get(fieldIndexInStruct).name();
+          int indexFromStructFieldName = Integer.valueOf(structFieldName.substring(5));
+          if (actualTypeIndex == indexFromStructFieldName) {
+            relatedFieldInStructFound = true;
+            options.add(visit(type.asStructType().fields().get(fieldIndexInStruct).type(), schema, visitor));
+            fieldIndexInStruct++;
+          }
+        }
+
+        // If a field is not projected, a corresponding field in the struct of Iceberg schema cannot be found
+        // for current type of union in Avro schema, a reader for current type still needs to be created and
+        // used to make the reading of Avro file successfully. In this case, a null field type is used to
+        // create the option for the reader of the current type which still can read the corresponding content
+        // in Avro file successfully.
+        if (!relatedFieldInStructFound) {
+          options.add(visit((Type) null, schema, visitor));
+        }
       }
       typeIndex++;
     }
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -20,7 +20,9 @@
 package org.apache.iceberg.avro;
 
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import org.apache.avro.Schema;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
@@ -28,6 +30,8 @@ import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
 
 public abstract class AvroSchemaWithTypeVisitor<T> {
+  private static final String UNION_TAG_FIELD_NAME = "tag";
+
   public static <T> T visit(org.apache.iceberg.Schema iSchema, Schema schema, AvroSchemaWithTypeVisitor<T> visitor) {
     return visit(iSchema.asStruct(), schema, visitor);
   }
@@ -97,13 +101,23 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
         options.add(visit(type, branch, visitor));
       }
     } else { // complex union case
-      int index = 1;
-      for (Schema branch : types) {
-        if (branch.getType() == Schema.Type.NULL) {
-          options.add(visit((Type) null, branch, visitor));
+      Map<Integer, Types.NestedField> icebergFieldMap = new HashMap<>();
+      for (Types.NestedField icebergField : type.asStructType().fields()) {
+        String fieldName = icebergField.name();
+        if (fieldName.equals(UNION_TAG_FIELD_NAME)) {
+          continue;
+        }
+        int fieldId = Integer.valueOf(fieldName.substring(5));
+        icebergFieldMap.put(fieldId, icebergField);
+      }
+
+      for (int i = 0; i < types.size(); ++i) {
+        Schema schema = types.get(i);
+        int fieldIdxInIceberg = types.get(0).getType() == Schema.Type.NULL ? i - 1 : i;
+        if (schema.getType() == Schema.Type.NULL || !icebergFieldMap.containsKey(fieldIdxInIceberg)) {
+          options.add(visit((Type) null, schema, visitor));
         } else {
-          options.add(visit(type.asStructType().fields().get(index).type(), branch, visitor));
-          index += 1;
+          options.add(visit(icebergFieldMap.get(fieldIdxInIceberg).type(), schema, visitor));
         }
       }
     }
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -20,9 +20,7 @@
 package org.apache.iceberg.avro;
 
 import java.util.Deque;
-import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


wmoustafa:
Thank you so much for the PR and comprehensive description!

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,13 +101,23 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+        }
+        int fieldId = Integer.valueOf(fieldName.substring(5));
+        icebergFieldMap.put(fieldId, icebergField);
+      }
+
+      for (int i = 0; i < types.size(); ++i) {
+        Schema schema = types.get(i);
+        int fieldIdxInIceberg = types.get(0).getType() == Schema.Type.NULL ? i - 1 : i;
+        if (schema.getType() == Schema.Type.NULL || !icebergFieldMap.containsKey(fieldIdxInIceberg)) {
+          options.add(visit((Type) null, schema, visitor));
```
wmoustafa, line 156: Can we add a comment to the code explaining why we pass `null` as the type if the union branch is not projected (or found in the map)?

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,13 +101,23 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+        if (fieldName.equals(UNION_TAG_FIELD_NAME)) {
+          continue;
+        }
+        int fieldId = Integer.valueOf(fieldName.substring(5));
+        icebergFieldMap.put(fieldId, icebergField);
+      }
+
+      for (int i = 0; i < types.size(); ++i) {
+        Schema schema = types.get(i);
+        int fieldIdxInIceberg = types.get(0).getType() == Schema.Type.NULL ? i - 1 : i;
```
wmoustafa: `fieldIdxInIceberg` --> `InferredFieldIndex`

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,13 +101,23 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
-        if (branch.getType() == Schema.Type.NULL) {
-          options.add(visit((Type) null, branch, visitor));
+      Map<Integer, Types.NestedField> icebergFieldMap = new HashMap<>();
+      for (Types.NestedField icebergField : type.asStructType().fields()) {
+        String fieldName = icebergField.name();
+        if (fieldName.equals(UNION_TAG_FIELD_NAME)) {
+          continue;
+        }
+        int fieldId = Integer.valueOf(fieldName.substring(5));
+        icebergFieldMap.put(fieldId, icebergField);
```
wmoustafa: `icebergFieldMap` --> `structFieldsByIndex`

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,13 +101,23 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
         options.add(visit(type, branch, visitor));
       }
     } else { // complex union case
-      int index = 1;
-      for (Schema branch : types) {
-        if (branch.getType() == Schema.Type.NULL) {
-          options.add(visit((Type) null, branch, visitor));
+      Map<Integer, Types.NestedField> icebergFieldMap = new HashMap<>();
+      for (Types.NestedField icebergField : type.asStructType().fields()) {
```
wmoustafa: `icebergField` --> `structField`

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,13 +101,23 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+        String fieldName = icebergField.name();
+        if (fieldName.equals(UNION_TAG_FIELD_NAME)) {
+          continue;
+        }
+        int fieldId = Integer.valueOf(fieldName.substring(5));
```
wmoustafa: `fieldIndex`

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
wmoustafa: Please add a comment in the code explaining the inference logic.

wmoustafa: It seems that `NULL` cannot always be in the first branch? Also something is off if we keep checking on `types.get(0)` in a loop on `types`.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,13 +101,23 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
         options.add(visit(type, branch, visitor));
       }
     } else { // complex union case
-      int index = 1;
-      for (Schema branch : types) {
-        if (branch.getType() == Schema.Type.NULL) {
-          options.add(visit((Type) null, branch, visitor));
+      Map<Integer, Types.NestedField> icebergFieldMap = new HashMap<>();
```
wmoustafa: I think you could avoid the map and index bookkeeping if you iterate on both types in sort-merge style in just one loop. I will let you decide if this looks simpler.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -304,13 +306,28 @@ protected void set(InternalRow struct, int pos, Object value) {
...
+    private UnionReader(Schema schema, List<ValueReader<?>> readers, Type expected) {
       this.schema = schema;
       this.readers = new ValueReader[readers.size()];
       for (int i = 0; i < this.readers.length; i += 1) {
         this.readers[i] = readers.get(i);
       }
+      this.projectedFieldIdsToIdxInReturnedRow = new HashMap<>();
+      this.isTagFieldProjected = false;
+      int idxInReturnedRow = 0;
+      for (Types.NestedField icebergField : expected.asStructType().fields()) {
```
wmoustafa: I am not sure if the naming convention of `Iceberg` is followed in this file. Maybe we can avoid calling it `icebergField`?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -304,13 +306,28 @@ protected void set(InternalRow struct, int pos, Object value) {
...
+        String fieldName = icebergField.name();
+        if (fieldName.equals("tag")) {
```
wmoustafa: Can we use the constant here?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -304,13 +306,28 @@ protected void set(InternalRow struct, int pos, Object value) {
...
+          this.isTagFieldProjected = true;
+          idxInReturnedRow++;
+          continue;
+        }
+        int fieldId = Integer.valueOf(fieldName.substring(5));
+        this.projectedFieldIdsToIdxInReturnedRow.put(fieldId, idxInReturnedRow++);
```
wmoustafa: Does it make sense to use an array here instead?

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -334,22 +351,20 @@ public InternalRow read(Decoder decoder, Object reuse) throws IOException {
...
-        struct.update(index + 1, value);
-        struct.setInt(0, index);
-      } else if (index < nullIndex) {
-        struct.update(index + 1, value);
-        struct.setInt(0, index);
-      } else {
-        struct.update(index, value);
-        struct.setInt(0, index - 1);
+      if (projectedFieldIdsToIdxInReturnedRow.containsKey(fieldId)) {
+        struct.update(projectedFieldIdsToIdxInReturnedRow.get(fieldId), value);
```
wmoustafa: I am not sure if this line would impact performance since it uses a hashmap lookup. Maybe another reason to use an array.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: Done.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: Done.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: Done

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: Done

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: Good point. I refactor this part of code without the map for bookkeeping and create a new function named `visitComplexUnion` to contain these code.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
yiqiangin: Done.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
yiqiangin: Done.

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
yiqiangin: Good point. An integer array is used instead. Also some refactoring has been made for SparkUnionReader

spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
yiqiangin: Agree.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +99,65 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
-          options.add(visit(type.asStructType().fields().get(index).type(), branch, visitor));
-          index += 1;
+      visitComplexUnion(type, union, visitor, options);
+    }
+    return visitor.union(type, union, options);
+  }
+
+  /*
+  A complex union with multiple types of Avro schema is converted into a struct with multiple fields of Iceberg schema.
+  A field is related to a type in the order defined in Avro schema. Also, an extra tag field is added into the struct of
```
wmoustafa: Should we frame `A field is related to a type in the order defined in Avro schema` as an expectation from the Iceberg schema to match the Avro schema?

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +99,65 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+  Iceberg schema. The user can query the column of union type with the field projected (e.g. colUnion.field0) in which
```
wmoustafa: I think Iceberg documentation tries to avoid describing scenarios referencing the user. Can we make this a factual statement about the function spec that does not involve referencing users?

wmoustafa: I am not sure if we should make this function description revolve around projection pruning use case. Let us describe it in a more general sense.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +99,65 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+  the maximum number of the fields to be projected equals to the number of fields of the complete struct converted
+  from the union. The case of without field projection equals to the case of full fields projection.
```
wmoustafa: `The case of without field projection equals to the case of full fields projection.` is not clear.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +99,65 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+        options.add(visit((Type) null, schema, visitor));
+        typeIndex++;
+        continue;
+      }
+
+      // If a NULL type is found before current type, the type index is one larger than the actual type index which
+      // can be used to track the corresponding field in the struct of Iceberg schema.
+      int actualTypeIndex = nullTypeFound ? typeIndex - 1 : typeIndex;
+      boolean relatedFieldInStructFound = false;
+      while (fieldIndexInStruct < type.asStructType().fields().size()) {
```
wmoustafa: Do you need this loop or can you just keep two pointers and leverage the outer loop?

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +99,65 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+    int fieldIndexInStruct = 0;
+    while (typeIndex < union.getTypes().size()) {
+      Schema schema = union.getTypes().get(typeIndex);
+      // in some cases, a NULL type exists in the union of Avro schema besides the actual types,
+      // and it affects the index of the actual types of the order in the union
+      if (schema.getType() == Schema.Type.NULL) {
+        nullTypeFound = true;
+        options.add(visit((Type) null, schema, visitor));
+        typeIndex++;
+        continue;
```
wmoustafa: Is it possible to do `if/else` instead of `continue`? I think this might get rid of `typeIndex++` twice.

core/src/main/java/org/apache/iceberg/avro/Avro.java
```diff
@@ -446,6 +447,11 @@ public ReadBuilder classLoader(ClassLoader classLoader) {
       return this;
     }
 
+    public ReadBuilder setFileSchema(Schema fileSchema) {
```
wmoustafa, line 450: This is only called in the tests. I am not sure how it is called in the non-test path.

core/src/main/java/org/apache/iceberg/avro/ProjectionDatumReader.java
```diff
@@ -59,7 +71,9 @@ public void setRowPositionSupplier(Supplier<Long> posSupplier) {
 
   @Override
   public void setSchema(Schema newFileSchema) {
-    this.fileSchema = newFileSchema;
+    if (this.fileSchema == null) {
+      this.fileSchema = newFileSchema;
+    }
```
wmoustafa, line 76: It seems we are ignoring `newFileSchema` in the uniontype case, which might be an incorrect use of the API. Note that this method overrides its parent in Avro's `DatumReader`.

core/src/main/java/org/apache/iceberg/avro/Avro.java
yiqiangin: Yes, this code together with the code change in `ProjectionDatumReader.setSchema` are added only for test the case that `fileSchema` to read Avro file is different from the actual schema to write the Avro file. This test case proves that the data cannot be returned by mistake if the schema to read the file is different from the schema to write the file.  In normal code flow, the `fileSchema` to read Avro file is always read from the metadata of the actual Avro file, therefore the schema to read file is always the same as the schema to write the file.  This code change is for test only and does not affect the normal code flow. It can be removed.

core/src/main/java/org/apache/iceberg/avro/ProjectionDatumReader.java
yiqiangin: Pls refer to the comment above.

PR #109, opened 2022-05-19
Retain the type of field while copying the default values.

maluchari:
While building  a read schema to support custom readers the field types should be retained.  If default values are present in table schema the values should be copied over but without changing the type.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
+++ b/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
@@ -94,8 +94,11 @@ class BuildAvroProjection extends AvroCustomOrderSchemaVisitor<Schema, Schema.Fi
         // if the expectedField has a defaultValue, but the avroField does not, we need to
         // create a newField to copy over the non-null default value.
         if (expectedField.hasDefaultValue() && !AvroSchemaUtil.hasNonNullDefaultValue(avroField)) {
+          Schema newFiledSchema = (expectedField.isOptional()) ?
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java
```diff
@@ -92,13 +92,10 @@ public Schema record(Schema record, List<String> names, Iterable<Schema.Field> s
...
         // if the expectedField has a defaultValue, but the avroField does not, we need to
-        // create a newField to copy over the non-null default value
+        // create a newField to copy over the non-null default value.
         if (expectedField.hasDefaultValue() && !AvroSchemaUtil.hasNonNullDefaultValue(avroField)) {
-          Schema newFiledSchema = (expectedField.isOptional()) ?
-              AvroSchemaUtil.toOption(AvroSchemaUtil.convert(expectedField.type()), true) :
-              AvroSchemaUtil.convert(expectedField.type());
           Schema.Field newField =
-              new Schema.Field(avroField.name(), newFiledSchema, avroField.doc(), expectedField.getDefaultValue());
+              new Schema.Field(avroField.name(), avroField.schema(), avroField.doc(), expectedField.getDefaultValue());
```
rzhang10: If the `avroField` is nullable avro schema, and the default value here is not null, the constructed avro schema will be invalid, please still retain the logic of making `AvroSchemaUtil.toOption` with putting null the second element in the avro union.

rzhang10 (approved):
LGTM, thanks!

PR #112, opened 2022-06-04
Fix and refactor schema parser

rzhang10:
This patch fixes and refactors the `SchemaParser` class to make its Json serialization and deserialization of Avro based default value logic stable and correct.


maluchari (approved):
Thanks for the fix @rzhang10

PR #113, opened 2022-06-08
Enhance the UT for testing required fields with default values

maluchari:
Enhanced the existing UT to extend testing default values for array, map, fixed, nested struct and map of key to array of values.

### DIFF
--- a/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroReaderForFieldsWithDefaultValue.java
+++ b/spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroReaderForFieldsWithDefaultValue.java
@@ -172,29 +172,29 @@ public class TestSparkAvroReaderForFieldsWithDefaultValue {
     Assert.assertEquals(expected.size(), rows.size());
     for (int row = 0; row < expected.size(); row++) {
       InternalRow actualRow = rows.get(row);
-      final InternalRow expectedFirstRow = new GenericInternalRow(8);
-      expectedFirstRow.update(0, UTF8String.fromString((String) expected.get(row).get(0)));
-      expectedFirstRow.update(1, UTF8String.fromString("foo"));
-      expectedFirstRow.update(2, 1);
-      expectedFirstRow.update(3, new ArrayBasedMapData(
+      final InternalRow expectedRow = new GenericInternalRow(8);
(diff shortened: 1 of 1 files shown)


spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroReaderForFieldsWithDefaultValue.java
```diff
@@ -50,72 +52,145 @@ public class TestSparkAvroReaderForFieldsWithDefaultValue {
...
-            null, null), new org.apache.avro.Schema.Field(nullableFiledName,
-            org.apache.avro.Schema.createUnion(org.apache.avro.Schema.create(INT),
-                org.apache.avro.Schema.create(NULL)), null, defaultValue)));
+    String writeSchemaString = "{\n" +
+            "  \"namespace\": \"com.n1\",\n" +
+            "  \"type\": \"record\",\n" +
+            "  \"name\": \"n1\",\n" +
+            "  \"fields\": [\n" +
+            "    {\n" +
+            "      \"name\": \"f1\",\n" +
```
rzhang10: Can we use a field `f0` with string type (or whatever simple type) without default, and make this `f1` also testable in the subsequent read schema? The `string` primitive type should also be tested by this `f1` field.

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkAvroReaderForFieldsWithDefaultValue.java
```diff
@@ -50,72 +52,150 @@ public class TestSparkAvroReaderForFieldsWithDefaultValue {
...
-        if (i >= writeSchema.getFields().size() && fields.get(i).hasDefaultValue()) {
-          expectedValue = fields.get(i).getDefaultValue();
-        } else if (i < writeSchema.getFields().size()) {
-          expectedValue = expectedRow.get(i);
-        }
-        Type fieldType = fields.get(i).type();
-        Object actualValue = actualRow.isNullAt(i) ? null : actualRow.get(i, convert(fieldType));
-        Assert.assertEquals(expectedValue, actualValue);
-      }
+      final InternalRow expectedFirstRow = new GenericInternalRow(8);
```
rzhang10: One last comment, I think we should rename `expectedFirstRow` to just `expectedRow`.

yiqiangin (approved):

PR #114, opened 2022-06-15
Support single type union for ORC-vectorization reader

yiqiangin:
**Problem**
The class of `UnionConverter` in `VectorizedSparkOrcReaders` always treats the type of the input Iceberg schema as Struct and creates a `BaseOrcColumnVector` with a list of ColumnVector to multiple types.  However it is valid only in case of the complex union with multiple types. The type of the input iceberg schema is not Struct in case of single type union so that `UnionConverter` fails to be constructed in this case.

**Solution**
- The type of input Iceberg schema is stored as the original type in `UnionConverter` instead of converting to Struct type.
- Along with the type of input Iceberg schema, a list of `Converter` are passed to `UnionConverter`. The list is generated based on the types in the union by `OrcSchemaWithTypeVistor`. The number of `Converter` in the list equals to the number of types in the union. Therefore, if there is only one `Converter` in the list, it indicates that the union is a single type union, and the corresponding `ColumnVector` can be generated by the only `Converter` in the list.

### DIFF
--- a/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/VectorizedSparkOrcReaders.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/data/vectorized/VectorizedSparkOrcReaders.java
@@ -432,11 +432,11 @@ public class VectorizedSparkOrcReaders {
   }
 
   private static class UnionConverter implements Converter {
-    private final Types.StructType structType;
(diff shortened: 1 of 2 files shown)


rzhang10 (approved):
LGTM! thanks @yiqiangin for the clean and effective fix!

PR #115, opened 2022-07-11
PoC of the mapping from Avro field name to Iceberg field id

yiqiangin:
- The mapping from the field name in Avro schema to the field id in Iceberg schema is derived during the conversion from Avro schema to Iceberg schema in the function of `AvroSchemaUtil.convertToDeriveNameMapping` and the class of `SchemaToType`.
- The mapping of direct child fields of an Avro schema field is stored as a property named `AvroFieldNameToIcebergId` in this Avro schema field, therefore it can be easily accessed when Avro schema is traversed to generate the correspond readers to read Avro data file.
- In case of union, the key of the mapping is the name of the branch in the union.
- In case of complex union, the code of `AvroSchemaWithTypeVisitor.visitUnion()` first gets the mapping from the property of Avro schema, then get the field id in Iceberg schema using the type name in Avro schema, finally it uses the field id to get the field type in Iceberg schema:
  if the corresponding field in Iceberg schema exists, the field is used to create the reader together with Avro schema node; 
  if the field for the given field id does not exist in Iceberg schema (which means this field is not projected in Iceberg schema), a pseudo branch type is created based on the corresponding Avro schema node to faciltate the creation of the reader. 
- In the class of UnionReader, the rows read from Avro data file are filtered according to the fields existing in Iceberg schema.

### DIFF
--- a/spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
+++ b/spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
@@ -25,6 +25,7 @@ import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -34,6 +35,7 @@ import org.apache.avro.util.Utf8;
 import org.apache.iceberg.avro.ValueReader;
 import org.apache.iceberg.avro.ValueReaders;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
+import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.UUIDUtil;
 import org.apache.spark.sql.catalyst.InternalRow;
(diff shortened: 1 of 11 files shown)


spark/src/main/java/org/apache/iceberg/spark/data/SparkValueReaders.java
```diff
@@ -86,8 +88,8 @@ static ValueReader<InternalRow> struct(List<ValueReader<?>> readers, Types.Struc
     return new StructReader(readers, struct, idToConstant);
   }
 
-  static ValueReader<InternalRow> union(Schema schema, List<ValueReader<?>> readers) {
-    return new UnionReader(schema, readers);
+  static ValueReader<InternalRow> union(Schema schema, List<ValueReader<?>> readers, Type expected) {
+    return new UnionReader(schema, readers, expected);
```
wmoustafa, line 92: Could you add a comment on why `expected` is needed in the union case, but not all the other readers? It is an interesting change since it added `Type` as a dependency of this class, which was not the case before.

wmoustafa (commented):
Thanks for the PR. I think it is okay to file the PR upstream. I may have some further comments, but it should not block having this reviewed upstream. I would suggest adding more background to the description, and making it clearer since it seems it focuses more on the lower level details while skipping some of the bigger picture. Keeping low level details is good, but we should transition from a higher level as well.

PR #116, opened 2022-07-18
Patch upstream PR #3717 and its dependencies with some refactor

rzhang10:
This PR patches https://github.com/apache/iceberg/pull/3717 and its dependencies to make it work in LI-Iceberg 0.11. It will address our production issues where a commit succeeds but its metadata file is deleted (because Iceberg wrongly thinks it fails) and the table is corrupted.

A more detailed explanation could be found in the description of the upstream PR.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/TableProperties.java
+++ b/core/src/main/java/org/apache/iceberg/TableProperties.java
@@ -36,8 +36,17 @@ public class TableProperties {
   public static final String COMMIT_TOTAL_RETRY_TIME_MS = "commit.retry.total-timeout-ms";
   public static final int COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT = 1800000; // 30 minutes
 
-  public static final String COMMIT_NUM_STATUS_CHECKS = "commit.num-status-checks";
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/iceberg/TableProperties.java
```diff
@@ -36,9 +36,18 @@ private TableProperties() {
   public static final String COMMIT_TOTAL_RETRY_TIME_MS = "commit.retry.total-timeout-ms";
   public static final int COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT = 1800000; // 30 minutes
 
-  public static final String COMMIT_NUM_STATUS_CHECKS = "commit.num-status-checks";
+  public static final String COMMIT_NUM_STATUS_CHECKS = "commit.status-check.num-retries";
```
yiqiangin, line 39: minor: renaming it as something like "COMMIT_NUM_RETRIES_CHECKS", as the value of the string is changed.

yiqiangin (approved):
LGTM

core/src/main/java/org/apache/iceberg/TableProperties.java
rzhang10: The name is directly from OSS so I feel it's better to keep it aligned.

PR #117, opened 2022-07-25
Check for schema corruption and fix it on commit

jack-moseley:
Due to an issue that `client.getTable` seems to sometimes return corrupt tables (missing columns), adding a check where we compare the table columns to the columns from the `avro.schema.literal`, and if there is a mismatch, just set the columns directly to `avro.schema.literal`. Also include a log warning when this happens listing the columns to hopefully help with finding a root cause.

Tested with gobblin metadata ingestion and also added a unit test.

### DIFF
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
@@ -259,7 +259,7 @@ public class HiveMetadataPreservingTableOperations extends HiveTableOperations {
     }
   }
 
-  static List<FieldSchema> getColsFromAvroSchema(Schema schema)
+  private static List<FieldSchema> getColsFromAvroSchema(Schema schema)
       throws SerDeException {
     AvroObjectInspectorGenerator avroOI = new AvroObjectInspectorGenerator(schema);
     List<String> columnNames = avroOI.getColumnNames();
@@ -273,7 +273,7 @@ public class HiveMetadataPreservingTableOperations extends HiveTableOperations {
         .collect(Collectors.toList());
   }
 
-  static String getAvroSchemaLiteral(Table table) {
+  private static String getAvroSchemaLiteral(Table table) {
     String schemaStr = table.getParameters().get(AvroSerdeUtils.AvroTableProperties.SCHEMA_LITERAL.getPropName());
     if (Strings.isNullOrEmpty(schemaStr)) {
       schemaStr = table.getSd().getSerdeInfo().getParameters()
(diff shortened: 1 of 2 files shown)


hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
```diff
@@ -209,6 +221,67 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+    if (schemaMismatched) {
+      LOG.warn("Schema columns don't match avro.schema.literal, setting columns to avro.schema.literal. Schema " +
+              "columns: {}, avro.schema.literal columns: {}",
+          table.getSd().getCols().stream().map(Object::toString).collect(Collectors.joining(", ")),
+          hiveCols.stream().map(Object::toString).collect(Collectors.joining(", ")));
+      table.getSd().setCols(hiveCols);
+    }
+  }
+
+  static List<FieldSchema> getColsFromAvroSchema(Schema schema)
```
rzhang10: nit: Maybe make this private as it's a helper method?

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
```diff
@@ -209,6 +221,67 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+    if (columnNames.size() != columnTypes.size()) {
+      throw new IllegalStateException();
+    }
+
+    return IntStream.range(0, columnNames.size())
+        .mapToObj(i -> new FieldSchema(columnNames.get(i), columnTypes.get(i).getTypeName(), ""))
+        .collect(Collectors.toList());
+  }
+
+  static String getAvroSchemaLiteral(Table table) {
```
rzhang10: nit: Maybe make this private as it's a helper method?

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
```diff
@@ -209,6 +221,67 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+      throws SerDeException {
+    AvroObjectInspectorGenerator avroOI = new AvroObjectInspectorGenerator(schema);
+    List<String> columnNames = avroOI.getColumnNames();
+    List<TypeInfo> columnTypes = avroOI.getColumnTypes();
+    if (columnNames.size() != columnTypes.size()) {
+      throw new IllegalStateException();
+    }
+
+    return IntStream.range(0, columnNames.size())
+        .mapToObj(i -> new FieldSchema(columnNames.get(i), columnTypes.get(i).getTypeName(), ""))
```
rzhang10, line 272: Are we sure TypeInfo's type name representation can be used to construct `FieldSchema` ? In other words, are they using the same type name representation, even for nested types?

jack-moseley: Yes for nested types the type will be for example `struct<field1:string,field2:string>`. I updated the unit test to fix a mismatch on a nested field.

rzhang10 (approved):
LGTM

PR #118, opened 2022-07-27
ORC: Handle query where select and filter only uses default value col…

rzhang10:
…umns

When a table evolves it schema with adding a column with default value, say `<col_def: string>` with default `foo`, and subsequently a query like `select col_def from table where col_def = 'bar'` is issued, `ORCSchemaUtil.buildOrcProjection` will output an empty struct corresponding ORC schema, since the `col_def` is not in the previously written files at all.

In this case, we shouldn't then push down the `col_def = 'bar'` filter to ORC files via its `SearchArgument` API, otherwise Iceberg will throw an exception.

This kind of filter is handled at Iceberg code level itself.

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
@@ -97,9 +97,13 @@ class OrcIterable<T> extends CloseableGroup implements CloseableIterable<T> {
       fileSchemaWithIds = ORCSchemaUtil.applyNameMapping(fileSchema, nameMapping);
     }
     readOrcSchema = ORCSchemaUtil.buildOrcProjection(schema, fileSchemaWithIds);
+    // If the projected ORC schema is an empty struct, it means we are only projecting columns
+    // with default values that aren't existent in previously written files, and thus we won't need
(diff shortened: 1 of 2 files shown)


orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
```diff
@@ -97,9 +97,13 @@ public CloseableIterator<T> iterator() {
...
     }
     readOrcSchema = ORCSchemaUtil.buildOrcProjection(schema, fileSchemaWithIds);
+    // If the projected ORC schema is an empty struct, it means we are only projecting columns
+    // with default values that aren't existent in previously written files, and thus we won't need
+    // to push down filters to ORC's SearchArgument, since we are not reading anything from files at all
+    boolean isEmptyStruct = readOrcSchema.getChildren().size() == 0;
 
     SearchArgument sarg = null;
-    if (filter != null) {
+    if (filter != null && !isEmptyStruct) {
```
aastha25, line 106: can we add a UT validating this?

rzhang10: Sure, just added.

aastha25:
LGTM! thanks for resolving the issue

PR #119, opened 2022-08-08
Set ORC columns and fix case-sensitivity issue with schema check

jack-moseley:
- It seems that the corrupt schema issue can also affect ORC SD properties `columns` and `columns.types`, so those need to be set when fixing the schema as well
- The check between schema fields should be case insensitive, otherwise if `avro.schema.literal` contains uppercase, the schema will be detected as mismatched every time

### DIFF
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
@@ -36,6 +36,7 @@ import org.apache.hadoop.hive.metastore.TableType;
 import org.apache.hadoop.hive.metastore.api.EnvironmentContext;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
+import org.apache.hadoop.hive.metastore.api.SerDeInfo;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.serde2.SerDeException;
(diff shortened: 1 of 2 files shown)


hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
```diff
@@ -282,6 +291,15 @@ private static String getAvroSchemaLiteral(Table table) {
     return schemaStr;
   }
 
+  private static void updateORCStorageDesc(List<FieldSchema> hiveCols, Table table) {
+    String columnsString = hiveCols.stream().map(FieldSchema::getName).collect(Collectors.joining(","));
+    String typesString = hiveCols.stream().map(FieldSchema::getType).collect(Collectors.joining(","));
+
+    Map<String, String> sdParams = table.getSd().getSerdeInfo().getParameters();
```
rzhang10, line 307: This just creates an in-memo map the of the sdParams but you didn't persist them to HMS right? Can we add a test on this?

jack-moseley: No, it's getting a reference to the map of the sdParams, so it will set them correctly. Updated the unit test to confirm that.

I did notice though that it could cause NPE if serdeInfo or parameters are null, so added a null check for those.

rzhang10: Got it, your first point makes sense.
But if the serdeInfo or parameters are null, shouldn't we create a new map and set the map as the table's serdeInfo/parameters ?

rzhang10 (approved):
LGTM

PR #120, opened 2022-08-10
Fix MergeHiveSchemaWithAvro to make it copy full Avro schema attributes

rzhang10:
The MergeHiveSchemaWithAvro merges a Hive Schema (input 1) and a corresponding Avro schema (input 2) to derive a more information-rich Avro schema.

For Avro map and array types, the existing logic doesn't copy over the attribute metadata associated with input 2 into the final output, resulting in metadata information loss in the final result Avro schema.

This patch fix that by copying over the metadata for those 2 types. This will ensure further Avro -> Iceberg conversions are correct (since Iceberg depends on Avro id metadata such as key-id, value-id and element-id) .

### DIFF
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
@@ -21,6 +21,7 @@ package org.apache.iceberg.hivelink.core.schema;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.avro.JsonProperties;
 import org.apache.avro.LogicalTypes;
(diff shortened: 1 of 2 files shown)


hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
```diff
@@ -113,6 +114,11 @@ public Schema list(ListTypeInfo list, Schema partner, Schema elementResult) {
     // if there was no matching Avro list, or if matching Avro list was an option, return an optional list
     boolean shouldResultBeOptional = partner == null || AvroSchemaUtil.isOptionSchema(partner);
     Schema result = Schema.createArray(elementResult);
+    if (partner != null) {
```
yiqiangin: Minor: can these code be abstracted into a function, so they are not duplicated in two places.

yiqiangin (approved):

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
rzhang10: good call! I'll update the code.

PR #121, opened 2022-08-12
Hive: Return null for currentSnapshot()

shardulm94:
We encounter an error when fetching properties for Hive tables read through Spark

```
java.lang.UnsupportedOperationException: Snapshots not supported for Hive tables without Iceberg metadata
  at xxx_shaded.org.apache.iceberg.hivelink.core.LegacyHiveTable.currentSnapshot(LegacyHiveTable.java:121)
  at xxx_shaded.org.apache.iceberg.spark.source.SparkTable.properties(SparkTable.java:138)
  ... 85 elided
```

This is because `SparkTable.properties()` tries to [fetch currentSnapshot()](https://github.com/linkedin/iceberg/blob/ddba6efc82034358a5e1bcc2502b36264d05a40c/spark3/src/main/java/org/apache/iceberg/spark/source/SparkTable.java#L138) which is not supported in `LegacyHiveTable`.

I propose here to change the behavior of currentSnapshot() to return a null and hence signal that no snapshot exists. This is also what vanilla Iceberg tables do and can be seen from how its handled at these examples [1](https://github.com/linkedin/iceberg/blob/ddba6efc82034358a5e1bcc2502b36264d05a40c/spark3/src/main/java/org/apache/iceberg/spark/source/SparkTable.java#L138) [2](https://github.com/linkedin/iceberg/blob/ddba6efc82034358a5e1bcc2502b36264d05a40c/core/src/main/java/org/apache/iceberg/IncrementalDataTableScan.java#L66) [3](https://github.com/linkedin/iceberg/blob/ddba6efc82034358a5e1bcc2502b36264d05a40c/spark2/src/main/java/org/apache/iceberg/spark/source/Reader.java#L309)

Unfortunately does not seem like there is an easy way to write a unit test for this as the bug is encountered during Spark scan of such a table, but the hivelink-core module is not set up to run Spark datasource tests.

### DIFF
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTable.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTable.java
@@ -118,7 +118,7 @@ public class LegacyHiveTable implements Table, HasTableOperations {
 
   @Override
   public Snapshot currentSnapshot() {
-    throw new UnsupportedOperationException("Snapshots not supported for Hive tables without Iceberg metadata");
+    return null;
   }
 
   @Override
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTable.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTable.java
@@ -19,6 +19,7 @@
 
 package org.apache.iceberg.hivelink.core;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import org.apache.iceberg.AppendFiles;
@@ -128,7 +129,7 @@ public class LegacyHiveTable implements Table, HasTableOperations {
 
   @Override
   public Iterable<Snapshot> snapshots() {
-    throw new UnsupportedOperationException("Snapshots not supported for Hive tables without Iceberg metadata");
+    return Collections.emptyList();
   }
 
   @Override


rzhang10:
@shardulm94 I don't suggest doing change this way. Our goal forward is to not to introduce linkedin specific changes to OSS iceberg code files, and only do those in the hivelink module.

Can you just let `LegacyHiveTable.currentSnapshot` return `null` instead? Does that work for this scenario?

shardulm94:
> @shardulm94 I don't suggest doing change this way. Our goal forward is to not to introduce linkedin specific changes to OSS iceberg code files, and only do those in the hivelink module.
> 
> Can you just let `LegacyHiveTable.currentSnapshot` return `null` instead? Does that work for this scenario?

Yes, that was my first solution, see [Hive: Return null for currentSnapshot()](https://github.com/linkedin/iceberg/pull/121/commits/84fb5c9cc2f1f823393dc11bb42667a1d087437e) but felt like this was safer. But if returning null better fits the policy, then I can use that.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/LegacyHiveTable.java
```diff
@@ -118,7 +118,7 @@ public String location() {
 
   @Override
   public Snapshot currentSnapshot() {
-    throw new UnsupportedOperationException("Snapshots not supported for Hive tables without Iceberg metadata");
+    return null;
```
rzhang10, line 122: Maybe for consistency purposes, also return `null` for `public Snapshot snapshot(long snapshotId)` and return an empty list for `public Iterable<Snapshot> snapshots()` ?

shardulm94: IMO, returning null for `public Snapshot snapshot(long snapshotId)` would not be correct. User is requesting a specific snapshotId and so it is a clear user error. WDYT?

I will change `snapshots()`

rzhang10: Sure, I'm fine with this.

shardulm94: Done

rzhang10 (approved):

PR #125, opened 2022-09-22
Bug fix: MergeHiveSchemaWithAvro should retain avro properties for li…

rzhang10:
…st and map when they are nullable.

We found a production issue when a ORC table with avro.schema contains a nullable array or map, the parallel tree walker `MergeHiveSchemaWithAvro` fails to copy over the persisted iceberg id properties in the schema literal, making the final output avro schema literal not have those properties, thus the final output avro schema conversion to Iceberg schema will fail.

This patch will fix this issue, and unblock production tables with such schema.


aastha25:
LGTM

yiqiangin (approved):

PR #132, opened 2022-12-07
[LI] Ignore field-ids from file schema in ORC read path via Hivelink

rzhang10:
Gobblin uses spark iceberg writer to write files and reuse those files for a hive table. In order for Iceberg reader to be able to read the Hive table via hivelink, we implement a runtime custom property in Iceberg Schema called `isFromHiveTable` to signal that this table is runtime converted from a Hive table, thus when the ORC read path construct a reader, it will ignore honoring the field-ids in the file schema and directly use `nameMapping` to map fields, which is what we need for the read to succeed.

Unit test in this case will be hard to write because it will need to mimic the entire Gobblin workflow to derive a table that satisfy the testing condition, so test will be done via integration test on gateway spark.


rzhang10:
Close this since this solution does not work from doing integration test.

PR #134, opened 2022-12-13
[LI] Add feature for Spark ORC reader to ignore field ids in files by using a new table property

rzhang10:
Adds a new table property `"read.orc.ignore.field-ids.enabled"` to control the Spark ORC reader behavior to ignore field-ids in file schema even if it contains it. This feature will be useful for LI-Iceberg to read Gobblin dual hive/Iceberg tables with shared iceberg written files.

Integration tested via spark-shell on the cluster, with setting the table property
```ALTER TABLE xxx.xxx SET TBLPROPERTIES ('read.orc.ignore.field-ids.enabled' = 'true');```
makes the table readable.

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
@@ -91,7 +91,7 @@ class OrcIterable<T> extends CloseableGroup implements CloseableIterable<T> {
     TypeDescription fileSchema = orcFileReader.getSchema();
     final TypeDescription readOrcSchema;
     final TypeDescription fileSchemaWithIds;
-    if (ORCSchemaUtil.hasIds(fileSchema) && !ignoreFileFieldIds) {
+    if (!ignoreFileFieldIds && ORCSchemaUtil.hasIds(fileSchema)) {
       fileSchemaWithIds = fileSchema;
     } else {
       if (nameMapping == null) {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by rdsr** [comment]
@rzhang10 is it possible to add unit tests here?

**Comment by rzhang10** [comment]
> @rzhang10 is it possible to add unit tests here?

No, I feel it's quite hard to do, because it's hard inside Iceberg codebase to create a hive table with iceberg files..(for that I will need to repeat all the logic gobblin is currently doing customly in the unit test).

I think we can rely on integration testing on the cluster.

**Comment by rdsr** [comment]
> > @rzhang10 is it possible to add unit tests here?
> 
> No, I feel it's quite hard to do, because it's hard inside Iceberg codebase to create a hive table with iceberg files..(for that I will need to repeat all the logic gobblin is currently doing customly in the unit test).
> 
> I think we can rely on integration testing on the cluster.

Would it make sense to
-  write with Iceberg. 
-  Open file header and remove the field-ids
-  Register the file with Hive table
-  and Query?

If this is far too complex then we can probably leave it out.. your call @rzhang10

File: orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java [hunk 1]
```diff
@@ -88,7 +91,7 @@ public CloseableIterator<T> iterator() {
     TypeDescription fileSchema = orcFileReader.getSchema();
     final TypeDescription readOrcSchema;
     final TypeDescription fileSchemaWithIds;
-    if (ORCSchemaUtil.hasIds(fileSchema)) {
+    if (ORCSchemaUtil.hasIds(fileSchema) && !ignoreFileFieldIds) {
```
yiqiangin: Nit: placing !ignoreFileFieldIds as the first condition could save the time to execute ORCSchemaUtil.hasIds() in case ignoreFileFieldIds = true

yiqiangin (approved):
LGTM

orc/src/main/java/org/apache/iceberg/orc/OrcIterable.java
rzhang10: that's a very good catch, it's indeed improve the performance, will update!

PR #135, opened 2023-01-13
Flink: Configuration to plan single file per task

DEEPTHIKORAT:
Having a single whole file per task allows to extract column metadata stats that can be used by the iceberg connector.
Added `PLAN_SINGLE_WHOLE_FILE_PER_TASK` configuration to ScanContext to specify that.
More context https://docs.google.com/document/d/1QgogtL1NhGCNpci6rbMlxJgnEp8lY5XNDOitKbAAyuk/edit#heading=h.ms1hwzaxm9te

In this PR we only make changes needed in flink-iceberg connector to allow for custom assigner to work. The actual implementation of the custom event-ordered assigner based on Netflix's design is not merged to li-iceberg as it lacks necessary testing required to make it available to the company. The assigner will be maintained by CODA team until required level of testing is added.

### DIFF
--- a/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
+++ b/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
@@ -80,6 +80,9 @@ public class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Boolean> PLAN_SINGLE_WHOLE_FILE_PER_TASK =
+      ConfigOptions.key("plan-single-whole-file-per-task").booleanType().defaultValue(false);
+
   private final boolean caseSensitive;
   private final boolean exposeLocality;
   private final Long snapshotId;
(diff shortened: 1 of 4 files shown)


flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -80,6 +80,9 @@ public class ScanContext implements Serializable {
   private static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
       ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
 
+  private static final ConfigOption<Boolean> PLAN_SINGLE_WHOLE_FILE_PER_TASK =
```
rzhang10, line 83: So I see you introduced this new conf option to control this feature.

Could you put this conf `PLAN_SINGLE_WHOLE_FILE_PER_TASK` in the PR description? Also, how does this config get passed down to `ScanContext` from client code? is it through table properties?

DEEPTHIKORAT: Added to the PR description. The ScanContext will be created by the client code using the builder. Woops, missed adding  the corresponding  IcebergSource builder API. Updated the PR.

yiqiangin (approved):
LGTM

rzhang10:
Could we add a unit test for this?

DEEPTHIKORAT:
> 

Unfortunately I did not see any Unit-tests for these classes. Would you prefer I add one to test this specific config?

rzhang10 (approved):
LGTM

PR #140, opened 2023-03-01
Fix reordering avro nullable types in MergeHiveSchemaWithAvro

rzhang10:
This patch fixes a bug where the avro nullable type orders (the order of null and the other type's order) is wrongly flipped, resulting in read failure.

The fix is demonstrated by the new unit test `shouldNotReorderListElementType`

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
@@ -138,6 +138,15 @@ public class AvroSchemaUtil {
     return false;
   }
 
+  public static int getNullIndex(Schema schema) {
+    Preconditions.checkArgument(isOptionSchema(schema));
+    if (schema.getTypes().get(0).getType() == Schema.Type.NULL) {
+      return 0;
+    } else {
+      return 1;
+    }
+  }
+
   /**
    * This method decides whether a schema represents a single type union, i.e., a union that contains only one option
    *
--- a/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
+++ b/hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
@@ -26,6 +26,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.avro.JsonProperties;
 import org.apache.avro.LogicalTypes;
 import org.apache.avro.Schema;
+import org.apache.avro.Schema.Type;
 import org.apache.hadoop.hive.serde2.typeinfo.DecimalTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.ListTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.MapTypeInfo;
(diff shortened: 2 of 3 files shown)


aastha25:
LGTM, thanks for the PR @rzhang10.
Please link the JIRA ticket in the description, if available.

core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java
```diff
@@ -138,6 +138,15 @@ public static boolean isOptionSchema(Schema schema) {
     return false;
   }
 
+  public static int getNullIndex(Schema schema) {
```
shardulm94, line 141: `nullIndexForOptionSchema`

Iceberg does not use `get` in method names. nulls can also be present in complex schemas, so ensure that we add the `OptionSchema` constraint in the method name itself.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
```diff
@@ -97,12 +98,12 @@ public Schema.Field field(String name, TypeInfo field, Schema.Field partner, Sch
...
    */
   private Schema reorderOptionIfRequired(Schema schema, Object defaultValue) {
-    if (AvroSchemaUtil.isOptionSchema(schema) && defaultValue != null) {
-      boolean isNullFirstOption = schema.getTypes().get(0).getType() == Schema.Type.NULL;
-      if (isNullFirstOption && defaultValue.equals(JsonProperties.NULL_VALUE)) {
-        return schema;
+    if (AvroSchemaUtil.isOptionSchema(schema)) {
+      int nullIndex = AvroSchemaUtil.getNullIndex(schema);
+      if (defaultValue != null && !defaultValue.equals(JsonProperties.NULL_VALUE)) {
+        return Schema.createUnion(schema.getTypes().get(1 - nullIndex), Schema.create(Type.NULL));
```
shardulm94, line 104: Hey @rzhang10 , I am having a hard time reasoning about these changes. Can you please include a description of how these changes help fix the issue. Can you also update the Java doc of this method with the new changes?

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
```diff
@@ -97,12 +98,12 @@ public Schema.Field field(String name, TypeInfo field, Schema.Field partner, Sch
...
       } else {
-        return Schema.createUnion(schema.getTypes().get(1), schema.getTypes().get(0));
+        return Schema.createUnion(Schema.create(Type.NULL), schema.getTypes().get(1 - nullIndex));
```
shardulm94, line 106: If no default value is specified, why do we need to reorder. The schema is valid as-is. The original intention of the method was to reorder to match the provided data type to work around invalid defaults.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
shardulm94: The new test passes without this change, why is this needed?

shardulm94: Ok I understand why this is needed. Previously `reorderOptionIfRequired` only worked correctly when `schema` was an option schema who first element was null. This was fine because we never returned an option with null as the second element from the children visitors, but now we do. I think the description needs to be updated to reflect the new behavior.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/schema/MergeHiveSchemaWithAvro.java
```diff
@@ -143,10 +144,14 @@ public Schema union(UnionTypeInfo union, Schema partner, List<Schema> results) {
...
     boolean shouldResultBeOptional = partner == null || AvroSchemaUtil.isOptionSchema(partner);
+    int nullIndex = 0;
+    if (partner != null && AvroSchemaUtil.isOptionSchema(partner)) {
+      nullIndex = AvroSchemaUtil.getNullIndex(partner);
+    }
     Schema hivePrimitive = hivePrimitiveToAvro(primitive);
     // if there was no matching Avro primitive, use the Hive primitive
     Schema result = partner == null ? hivePrimitive : checkCompatibilityAndPromote(hivePrimitive, partner);
-    return shouldResultBeOptional ? AvroSchemaUtil.toOption(result) : result;
+    return shouldResultBeOptional ? AvroSchemaUtil.toOption(result, nullIndex == 1) : result;
```
shardulm94, line 154: This handling is incomplete, it only handles cases for primitives inside containers (e.g. list), but does not handle similar cases for nested lists, maps, structs, etc. E.g. this case will fail
```
  @Test
  public void shouldNotReorderListElementType1() {
    String hive = "struct<fa:array<array<int>>>";
    Schema avro =
            SchemaBuilder.record("r1")
                    .fields()
                    .name("fa")
                    .type()
                    .nullable()
                    .array()
                    .items()
                    .nullable()
                    .array()
                    .items()
                    .nullable()
                    .intType()
                    .arrayDefault(Collections.singletonList(Arrays.asList(1, 2, 3)))
                    .endRecord();

    assertSchema(avro, merge(hive, avro));
  }
  ```
  
  I think we need similar handling in other places where we use the `shouldResultBeOptional` pattern, e.g. in the `struct`, `list` and `map` methods. Might be good to refactor the common pattern into a separate method for reuse.

PR #141, opened 2023-03-03
[LI] Bundle hivelink to iceberg-runtime jar

rzhang10:
Hivelink module is needed by downstream LI consumers as a common dependency.

We should bundle it into `iceberg-runtime` so they can define a single runtime dependency from `iceberg-runtime` to get everything they need at runtime. `iceberg-runtime` automatically takes care of shading and dependency management related stuff so we don't need to additionally do that for hivelink.


zheliu2 (approved):

ljfgem (approved):
LGTM

PR #142, opened 2023-03-20
Revert "Remove hive-serde dependency from HiveMetadataPreservingTableOperations (#138)"

jack-moseley:
This reverts commit 5bfe975d3fcf5517416cef0aab4971e2943824e2.

Reverting #138 since converting the schema to iceberg and back causes it to lose unionType info, causing schema issues for some datasets.

### DIFF
--- a/hivelink-core/src/test/java/org/apache/iceberg/hivelink/core/TestHiveMetadataPreservingTableOperations.java
+++ b/hivelink-core/src/test/java/org/apache/iceberg/hivelink/core/TestHiveMetadataPreservingTableOperations.java
@@ -47,9 +47,9 @@ public class TestHiveMetadataPreservingTableOperations {
 
     long currentTimeMillis = System.currentTimeMillis();
     StorageDescriptor storageDescriptor = new StorageDescriptor();
-    FieldSchema field1 = new FieldSchema("name", "string", null);
-    FieldSchema field2 = new FieldSchema("id", "int", null);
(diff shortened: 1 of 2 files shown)


hivelink-core/src/test/java/org/apache/iceberg/hivelink/core/TestHiveMetadataPreservingTableOperations.java
```diff
@@ -47,9 +47,9 @@ public void testFixMismatchedSchema() {
 
     long currentTimeMillis = System.currentTimeMillis();
     StorageDescriptor storageDescriptor = new StorageDescriptor();
-    FieldSchema field1 = new FieldSchema("name", "string", null);
-    FieldSchema field2 = new FieldSchema("id", "int", null);
-    FieldSchema field3 = new FieldSchema("nested", "struct<field1:string,field2:string>", null);
+    FieldSchema field1 = new FieldSchema("name", "string", "");
```
rzhang10, line 50: Could you please explain a bit why we need to make changes to these comment strings? I'm losing the context a bit.

jack-moseley: I think it was just that converting avro->hive using iceberg ended up setting the comment to null, while doing it through the hive-serde library ended up setting it to empty string

rzhang10 (approved):

PR #149, opened 2023-07-18
Support to field filtering of complex union type in Avro

yiqiangin:
**Problem**
Currently field filtering does not work for complex union type in Avro. The root cause of the problem is as follows:

The current code assumes that the types inside a union from Avro schema should match the type fields inside a union struct from Iceberg schema when Avro union reader is created.
In case of column projection of union type, the current code only prune the schema of union in Iceberg schema with the projected fields, while the union of Avro schema still contains all the types. It results in the mismatch between Avro schema and Iceberg schema for the union in this case.
However, as all the contents of each data type in a union in Avro file should be read by Avro readers correctly no matter this data type is projected or not based on the decoding procedure of Avro file, all the types in a union from Avro schema are needed to create the corresponding type readers in AvroUnionReader even in case of column projection. Therefore the union in Avro schema cannot be pruned like what is done to the union struct in Iceberg schema.

**Solution**
In addition, the order of the types in a union in Iceberg schema can be identified from its field name like "field0".."fieldK". K is the index which can be used to match the order of the types in the union of Avro schema.

In the code of create the readers of all types in the union of Avro schema (namely AvroSchemaWithTypeVisitor.visitUnion), checking the fields of the struct corresponding union in Iceberg schema to create a map between the order index and the field type in Iceberg schema. When iterating through all the types in Avro schema, using the order index to check if the corresponding type exists in the map, if yes which means the field is projected, creating the option of creating the reader with the type in Iceberg schema, otherwise, creating the option with a pseudo Iceberg type converted from Avro schema. 
There is a special case when the Avro schema is RECORD, the property of `field-id` should be set in each field under the RECORD Avro schema. Therefore, a `NameMapping` is created based on both Avro schema and the converted Iceberg type by the code of  `AvroWithPartnerByStructureVisitor` and `NameMappingWithAvroSchema` from PR [Add Avro-assisted name mapping](https://github.com/apache/iceberg/pull/7392)

In the code of AvroUnionReader, Iceberg schema needs to be passed into it. The fields of the returned row should be constructed based on the fields in Iceberg schema not the types in Avro schema. If tag field is projected, one more field is added in the beginning of the row and updated with the index of the field in Avro file.

**Test**
All the test cases in TestSparkAvroUnions.java with a new test case writeAndValidateRequiredComplexUnionWithProjection

```
val df = spark.sql("select c1.field0 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.field0,c1.field1 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.tag,c1.field0 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.tag,c1.field0,c1.field1 from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.field1,c1.field0,c1.tag from u_yiqding.avro_union_table_test")

val df = spark.sql("select c1.field1,c1.field0 from u_yiqding.avro_union_table_test")
```

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -108,6 +108,9 @@ public abstract class AvroSchemaWithTypeVisitor<T> {
   /*
   A complex union with multiple types of Avro schema is converted into a struct with multiple fields of Iceberg schema.
   Also an extra tag field is added into the struct of Iceberg schema during the conversion.
+  Given an example of complex union in both Avro and Iceberg:
+  Avro schema: {"name":"unionCol","type":["int","string"]}
+  Iceberg schema:  struct<0: tag: required int, 1: field0: optional int, 2: field1: optional string>
   The fields in the struct of Iceberg schema are expected to be stored in the same order
   as the corresponding types in the union of Avro schema.
   Except the tag field, the fields in the struct of Iceberg schema are the same as the types in the union of Avro schema
--- a/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
+++ b/core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
@@ -22,12 +22,15 @@ package org.apache.iceberg.avro;
 import java.util.Deque;
 import java.util.List;
 import org.apache.avro.Schema;
+import org.apache.iceberg.mapping.MappedFields;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
 
 public abstract class AvroSchemaWithTypeVisitor<T> {
+  private static final String UNION_TAG_FIELD_NAME = "tag";
+
   public static <T> T visit(org.apache.iceberg.Schema iSchema, Schema schema, AvroSchemaWithTypeVisitor<T> visitor) {
     return visit(iSchema.asStruct(), schema, visitor);
   }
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +100,89 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+    Type iType = AvroSchemaUtil.convert(schema);
+    if (schema.getType().equals(Schema.Type.RECORD)) {
+      // When the type of Avro schema is RECORD, the fields under it must have the property of "field-id".
+      // However, the "field-id" is not set in previous steps as the corresponding Iceberg type is not projected
+      // and no field id can be found for this field in Iceberg schema.
+      // Therefore, a name mapping is created based on the Avro schema and its corresponding Iceberg type.
+      // The field-id from the resulted name mapping is assigned as the property of "field-id" of each
+      // field under the Avro schema.
+      NameMappingWithAvroSchema nameMappingWithAvroSchema = new NameMappingWithAvroSchema();
+      MappedFields nameMapping = AvroWithPartnerByStructureVisitor.visit(
```
aastha25, line 179: do field-ids need to be unique across all the (top level & nested) fields in the schema? if yes, does this visitor achieve that?

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
```diff
@@ -97,17 +100,89 @@ private static <T> T visitUnion(Type type, Schema union, AvroSchemaWithTypeVisit
...
+                UNION_TAG_FIELD_NAME.equals(struct.fields().get(fieldIndexInStruct).name())) {
+          fieldIndexInStruct++;
+        }
+
+        if (fieldIndexInStruct < struct.fields().size()) {
+          // If a NULL type is found before current type, the type index is one larger than the actual type index which
+          // can be used to track the corresponding field in the struct of Iceberg schema.
+          int actualTypeIndex = nullTypeFound ? typeIndex - 1 : typeIndex;
+          String structFieldName = type.asStructType().fields().get(fieldIndexInStruct).name();
+          int indexFromStructFieldName = Integer.valueOf(structFieldName.substring(5));
```
aastha25, line 146: nit: I didnt realise that the `structFieldName` starts with the index. Could you elaborate the method doc to add an example of complex union schema and corresponding iceberg  type. It'll also help clarify that the iceberg type has 'tag' and doesnt have 'null'.

core/src/main/java/org/apache/iceberg/avro/AvroWithTypeByStructureVisitor.java
```diff
@@ -0,0 +1,62 @@
...
+ * under the License.
+ */
+
+package org.apache.iceberg.avro;
+
+import org.apache.iceberg.types.Type;
+import org.apache.iceberg.types.Types;
+import org.apache.iceberg.util.Pair;
+
+public class AvroWithTypeByStructureVisitor<T> extends AvroWithPartnerByStructureVisitor<Type, T> {
```
aastha25, line 32: could you add a javadoc for `AvroWithTypeByStructureVisitor` & `NameMappingWithAvroSchema`

core/src/main/java/org/apache/iceberg/avro/SchemaToType.java
```diff
@@ -121,7 +120,8 @@ public Type union(Schema union, List<Type> options) {
       return options.get(0);
     } else {
       // Complex union
-      List<Types.NestedField> newFields = new ArrayList<>();
+      // List<Types.NestedField> newFields = new ArrayList<>();
```
aastha25: nit: remove commented out line

aastha25 (approved):
thanks for the PR @yiqiangin ! left minor comments

core/src/main/java/org/apache/iceberg/avro/AvroWithTypeByStructureVisitor.java
yiqiangin: added

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: yes

core/src/main/java/org/apache/iceberg/avro/AvroSchemaWithTypeVisitor.java
yiqiangin: added

PR #150, opened 2023-10-06
Support to field filtering of complex union type in ORC

yiqiangin:
**Problem**
Currently the query on the whole column of complex union type works in case that the data are stored in ORC file, (e.g. the query like `select c1 from tbl` where c1 is a type of complex union, however, field filtering on the column of complex union in the query does not work (e.g the query like `select c1.field0 from tbl` where field0 is one type in the complex union). The root cause of the problem is as follows:
Two types of schema are used in Iceberg to read the data from ORC file per the columns defined in the user query: one is table schema from table metastore and used by the user query which is defined in Iceberg format, namely Iceberg schema; the other is file schema which is defined in ORC file and data are stored complying with this schema, namely ORC schema.  As Iceberg does not support union type natively, complex union is defined as a struct type in Iceberg schema. The current code assumes that the types inside a union from ORC schema should match the type fields inside a union struct from Iceberg schema when ORC union reader is created.
In case of field filtering of union type, the current code only prune the schema of union in Iceberg schema with the projected fields, while the union of ORC schema still contains all the types. It results in the mismatch between ORC schema and Iceberg schema for the union in this case.
However, as all the contents of each data type in a union in Orc file should be read by Orc readers correctly no matter this data type is projected or not based on the decoding procedure of Orc file, all the types in a union from Orc schema are needed to create the corresponding type readers in `OrcUnionReader` even in case of column projection. Therefore the union in Orc schema cannot be pruned like what is done to the union struct in Iceberg schema.

**Solution**
In addition, the order of the types in a union in Iceberg schema can be identified from its field name like "field0".."fieldK". K is the index which can be used to match the order of the types in the union of Orc schema.

In the code of create the readers of all types in the union of Orc schema (namely `OrcSchemaWithTypeVisitor.visitUnion`), checking the fields of the struct corresponding union in Iceberg schema to create a map between the order index and the field type in Iceberg schema. When iterating through all the types in Orc schema, using the order index to check if the corresponding type exists in the map, if yes which means the field is projected, creating the option of creating the reader with the type in Iceberg schema, otherwise, creating the option with a pseudo Iceberg type converted from Orc schema.

In the code of `OrcUnionReader`, Iceberg schema needs to be passed into it. The fields of the returned row should be constructed based on the fields in Iceberg schema not the types in Orc schema. If tag field is projected, one more field is added in the beginning of the row and updated with the index of the field in Orc file.

**Test**
All the test cases in `TestSparkOrcUnions.java` with a new test case `testComplexUnionWithColumnProjection`

```
val df = spark.sql("select c1.field0 from u_yiqding.orc_union_table_test")

val df = spark.sql("select c1.field0,c1.field1 from u_yiqding.orc_union_table_test")

val df = spark.sql("select c1.tag,c1.field0 from u_yiqding.orc_union_table_test")

val df = spark.sql("select c1.tag,c1.field0,c1.field1 from u_yiqding.orc_union_table_test")

val df = spark.sql("select c1.field1,c1.field0,c1.tag from u_yiqding.orc_union_table_test")

val df = spark.sql("select c1.field1,c1.field0 from u_yiqding.orc_union_table_test")
```

The schema of the table `u_yiqding.orc_union_table_test` is as follows:
```
# col_name            	data_type           	comment

c1                  	uniontype<int,string>
```

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -82,6 +82,7 @@ public final class ORCSchemaUtil {
   static final String ICEBERG_FIELD_LENGTH = "iceberg.length";
 
   public static final String ICEBERG_UNION_TAG_FIELD_NAME = "tag";
+  public static final int ICEBERG_UNION_TYPE_FIELD_NAME_PREFIX_LENGTH = 5;
   private static final ImmutableMultimap<Type.TypeID, TypeDescription.Category> TYPE_MAPPING =
       ImmutableMultimap.<Type.TypeID, TypeDescription.Category>builder()
           .put(Type.TypeID.BOOLEAN, TypeDescription.Category.BOOLEAN)
@@ -410,13 +411,15 @@ public final class ORCSchemaUtil {
                                                                  Map<Integer, OrcField> mapping,
                                                                  OrcField orcField) {
     TypeDescription orcType = TypeDescription.createUnion();
-    List<Types.NestedField> nestedFields = type.asStructType().fields();
+    List<Types.NestedField> nestedFieldsFromIcebergSchema = type.asStructType().fields();
     for (int i = 0; i < orcField.type.getChildren().size(); ++i) {
       TypeDescription childOrcType = orcField.type.getChildren().get(i);
       boolean typeProjectedInIcebergSchema = false;
-      for (Types.NestedField nestedField : nestedFields) {
+      for (Types.NestedField nestedField : nestedFieldsFromIcebergSchema) {
+        // the name of the field in the struct of Iceberg schema is in the pattern of "fieldx"
+        // where x is an integer index starting from 0
         if (!nestedField.name().equals(ICEBERG_UNION_TAG_FIELD_NAME) &&
-                Integer.parseInt(nestedField.name().substring(5)) == i) {
+                Integer.parseInt(nestedField.name().substring(ICEBERG_UNION_TYPE_FIELD_NAME_PREFIX_LENGTH)) == i) {
           // child type is projected in Iceberg schema
           TypeDescription childType = buildOrcProjection(nestedField.fieldId(), nestedField.type(),
                   isRequired && nestedField.isRequired(), mapping);
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -415,7 +415,7 @@ public final class ORCSchemaUtil {
       TypeDescription childOrcType = orcField.type.getChildren().get(i);
       boolean typeProjectedInIcebergSchema = false;
       for (Types.NestedField nestedField : nestedFields) {
-        if (!nestedField.name().equals("tag") &&
+        if (!nestedField.name().equals(ICEBERG_UNION_TAG_FIELD_NAME) &&
                 Integer.parseInt(nestedField.name().substring(5)) == i) {
           // child type is projected in Iceberg schema
           TypeDescription childType = buildOrcProjection(nestedField.fieldId(), nestedField.type(),
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -81,6 +81,7 @@ public final class ORCSchemaUtil {
   public static final String ICEBERG_LONG_TYPE_ATTRIBUTE = "iceberg.long-type";
   static final String ICEBERG_FIELD_LENGTH = "iceberg.length";
 
+  public static final String ICEBERG_UNION_TAG_FIELD_NAME = "tag";
   private static final ImmutableMultimap<Type.TypeID, TypeDescription.Category> TYPE_MAPPING =
       ImmutableMultimap.<Type.TypeID, TypeDescription.Category>builder()
           .put(Type.TypeID.BOOLEAN, TypeDescription.Category.BOOLEAN)
@@ -381,9 +382,9 @@ public final class ORCSchemaUtil {
 
   private static TypeDescription getOrcSchemaForUnionType(Type type, boolean isRequired, Map<Integer, OrcField> mapping,
       OrcField orcField) {
-    TypeDescription orcType;
+
     if (orcField.type.getChildren().size() == 1) { // single type union
-      orcType = TypeDescription.createUnion();
+      TypeDescription orcType = TypeDescription.createUnion();
 
       TypeDescription childOrcStructType = TypeDescription.createStruct();
       for (Types.NestedField nestedField : type.asStructType().fields()) {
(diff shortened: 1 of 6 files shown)


aastha25:
@yiqiangin thanks for the PR. 
Could you add some more details in the description and examples in the description

(1) what is currently supported prior to this PR, for example - I think only `spark.sql("select c1 from u_yiqding.orc_union_table_test")`? does this mean reads only work if the ORC file has values for all the types inside the uniontype? Also, please add the schema of `u_yiqding.orc_union_table_test` in the description.

(2) "types inside a union from Orc schema should match the type fields inside a union struct from Iceberg schema when Orc union reader is created."
the latter is constructed from the ORC data files and the former is constructed using ORC file schema / catalog schema?

yiqiangin:
>(1) what is currently supported prior to this PR, for example - I think only `spark.sql("select c1 from u_yiqding.orc_union_table_test")`? does this mean reads only work if the ORC file has values for all the types inside the uniontype? Also, please add the schema of `u_yiqding.orc_union_table_test` in the description.

Yes, currently it only works querying the whole column of complex union like `spark.sql("select c1 from u_yiqding.orc_union_table_test"). It does not require ORC file has values for all the types in the union. For each row, the column of complex union could only have value for one type. The returned result of the query has null value for the type which does not have valid value in the row. I have added this explaination in the description.

> (2) "types inside a union from Orc schema should match the type fields inside a union struct from Iceberg schema when Orc union reader is created." the latter is constructed from the ORC data files and the former is constructed using ORC file schema / catalog schema?

Two types of schema are used in Iceberg to read the data from ORC file per the columns defined in the user query: one is table schema from table metastore and used by the user query which is defined in Iceberg format, namely Iceberg schema; the other is file schema which is defined in ORC file and data are stored complying with this schema, namely ORC schema. As Iceberg does not support union type natively, complex union is defined as a struct type in Iceberg schema.
This explanation has been added in the description as well.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -399,13 +400,33 @@ private static TypeDescription getOrcSchemaForUnionType(Type type, boolean isReq
...
+                                                                 Map<Integer, OrcField> mapping,
+                                                                 OrcField orcField) {
+    TypeDescription orcType = TypeDescription.createUnion();
+    List<Types.NestedField> nestedFields = type.asStructType().fields();
+    for (int i = 0; i < orcField.type.getChildren().size(); ++i) {
+      TypeDescription childOrcType = orcField.type.getChildren().get(i);
+      boolean typeProjectedInIcebergSchema = false;
+      for (Types.NestedField nestedField : nestedFields) {
+        if (!nestedField.name().equals(ICEBERG_UNION_TAG_FIELD_NAME) &&
+                Integer.parseInt(nestedField.name().substring(5)) == i) {
```
aastha25: can we define a constant for '5' here and add a 1 liner explanation. Probably just an example would be fine to demonstrate how fieldname is extracted.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -399,13 +400,33 @@ private static TypeDescription getOrcSchemaForUnionType(Type type, boolean isReq
...
-        orcType.addUnionChild(childType);
+      return getOrcSchemaForComplexUnionType(type, isRequired, mapping, orcField);
+    }
+  }
+
+  private static TypeDescription getOrcSchemaForComplexUnionType(Type type, boolean isRequired,
+                                                                 Map<Integer, OrcField> mapping,
+                                                                 OrcField orcField) {
+    TypeDescription orcType = TypeDescription.createUnion();
+    List<Types.NestedField> nestedFields = type.asStructType().fields();
```
aastha25: general comment on this PR: Its a little hard to follow the OrcType and IcebergType fields & values. For example: here can we rename `nestedFields` to `nestedFieldsFromIcebergSchema`  to easily identify the reference.

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
```diff
@@ -78,14 +79,73 @@ protected T visitUnion(Type type, TypeDescription union, OrcSchemaWithTypeVisito
...
     }
 
     return visitor.union(type, union, options);
   }
 
+  /*
+  A complex union with multiple types of Orc schema is converted into a struct with multiple fields of Iceberg schema.
+  Also an extra tag field is added into the struct of Iceberg schema during the conversion.
+  Given an example of complex union in both Orc and Iceberg:
+  Orc schema: {"name":"unionCol","type":["int","string"]}
```
aastha25, line 93: Lets add a 1 liner clarifying that 'nulls' and 'defaults' are not a consideration for ORC schemas.

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
```diff
@@ -78,14 +79,73 @@ protected T visitUnion(Type type, TypeDescription union, OrcSchemaWithTypeVisito
...
+  // used to make the reading of Orc file successfully. In this case, a pseudo Iceberg type is converted from
+  // the Orc schema and is used to create the option for the reader of the current type which still can
+  // read the corresponding content in Orc file successfully.
+  private static <T> void visitNotProjectedTypeInComplexUnion(TypeDescription schema,
+                                                              OrcSchemaWithTypeVisitor<T> visitor,
+                                                              List<T> options,
+                                                              int typeIndex) {
+    OrcToIcebergVisitor schemaConverter = new OrcToIcebergVisitor();
+    schemaConverter.beforeField("field" + typeIndex, schema);
+    schema.setAttribute(org.apache.iceberg.orc.ORCSchemaUtil.ICEBERG_ID_ATTRIBUTE, "-1");
```
aastha25: this -1 can also be created as an 'invalid field id" constant. That would also clarify the objective - multiple fields can have an invalid field id.

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcValueReaders.java
```diff
@@ -166,12 +169,39 @@ protected void set(InternalRow struct, int pos, Object value) {
...
       this.readers = new OrcValueReader[readers.size()];
       for (int i = 0; i < this.readers.length; i += 1) {
         this.readers[i] = readers.get(i);
       }
+      this.expected = expected;
+
+      if (this.readers.length > 1) {
+        // Creating an integer array to track the mapping between the index of fields to be projected
+        // and the index of the value for the field stored in the returned row,
+        // if the value for a field equals to -1, it means the value of this field should not be stored
```
aastha25: can we use something apart from '-1' since we use that constant as a fieldId for invalid dummy fields. 
INT_MIN?

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcValueReaders.java
```diff
@@ -166,12 +169,39 @@ protected void set(InternalRow struct, int pos, Object value) {
 
   static class UnionReader implements OrcValueReader<Object> {
     private final OrcValueReader[] readers;
+    private final Type expected;
```
aastha25: 'expected' is derived from the icebergSchema. Can we rename it to 'expectedIcebergSchema'.

aastha25 (approved):
Thanks for working on the PR @yiqiangin . I have left minor doc comments.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
yiqiangin: Good point, done

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
yiqiangin: Good point, done

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
yiqiangin: Good point, done

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
yiqiangin: Good point, done

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcValueReaders.java
yiqiangin: Good point, done

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcValueReaders.java
yiqiangin: Good point, done

aastha25 (approved):

PR #151, opened 2023-11-30
Fix NPE excpetion when reading dataset with nested array

zheliu2:
## Context:
We have identified a issue when reading the schema of Iceberg tables. Specifically, if the table contains non-primitive Avro types (referred to as `TypeA`) as the first field and arrays of `TypeA` as the second field, it leads to a NullPointerException during loading the table schema.

## Root Cause
The root cause lies in the absence of handling default null values within nested objects when two fields inside a record is cross referenced, it is causing a failure in the condition check in Jackson codebases below:

```java
if (datum == JsonProperties.NULL_VALUE) {
    generator.writeNull();
}
```
When datum is Java `null`, this predicate will miss and resulting in unidentified types when Jackson is parsing the schema.

Below is the detailed stacktracestacktrace of the issue
```
java.lang.NullPointerException
	at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:96)
	at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:68)
	at org.apache.avro.util.internal.JacksonUtils.toJsonNode(JacksonUtils.java:53)
	at org.apache.avro.Schema$Field.<init>(Schema.java:586)
	at org.apache.iceberg.avro.TypeToSchema.struct(TypeToSchema.java:108)
	at org.apache.iceberg.avro.TypeToSchema.struct(TypeToSchema.java:36)
	at org.apache.iceberg.types.TypeUtil.visit(TypeUtil.java:349)
	at org.apache.iceberg.types.TypeUtil.visit(TypeUtil.java:358)
```

## Resolution
This pull request addresses the issue by ensuring that default null values in nested objects are converted to `JsonProperties.NULL_VALUE` when iceberg is reading the schema object, and convert the  `JsonProperties.NULL_VALUE`  to java default null when serializing the schema object. This prevents the NPE and subsequent failures during reading and writing the table.

Below is the pseudo-code of the implementation of the fucntion.

```java
Object convertToJsonNull(defaultValue):
    if defaultValue is a Map:
        for each entry in defaultValue:
            if entry value is a Map or Collection:
                entry value = convertToJsonNull(entry value)
            else:
                entry value = JsonProperties.NULL_VALUE
        return defaultValue

    else if defaultValue is a List:
        originalList = copy of defaultValue
        copiedList = new ArrayList
        for each element in originalList:
            if element is a Map or Collection:
                copiedList.add(convertToJsonNull(element))
            else:
                copiedList.add(JsonProperties.NULL_VALUE)
        return copiedList

    else:
        return defaultValue
```

## Test
1. All unit test passed
2. Created a unit that reproduces the original issue and it was fixed with the newly introduced function.

### DIFF
--- a/core/src/main/java/org/apache/iceberg/avro/TypeToSchema.java
+++ b/core/src/main/java/org/apache/iceberg/avro/TypeToSchema.java
@@ -19,6 +19,8 @@
 
 package org.apache.iceberg.avro;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
@@ -103,7 +105,8 @@ class TypeToSchema extends TypeUtil.SchemaVisitor<Schema> {
       String fieldName =  isValidFieldName ? origFieldName : AvroSchemaUtil.sanitize(origFieldName);
       Object defaultValue = structField.hasDefaultValue() ? structField.getDefaultValue() :
           (structField.isOptional() ? JsonProperties.NULL_VALUE : null);
-      Schema.Field field = new Schema.Field(fieldName, fieldSchemas.get(i), structField.doc(), defaultValue);
+      Schema.Field field = new Schema.Field(fieldName, fieldSchemas.get(i), structField.doc(),
+          convertToJsonNull(defaultValue));
       if (!isValidFieldName) {
         field.addProp(AvroSchemaUtil.ICEBERG_FIELD_NAME_PROP, origFieldName);
       }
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/iceberg/avro/TypeToSchema.java
```diff
@@ -232,4 +235,32 @@ public Schema primitive(Type.PrimitiveType primitive) {
...
   }
+
+  // This function ensures that all nested null are converted to JsonProperties.NULL_VALUE
+  // to make sure JacksonUtils.toJsonNode() converts them properly.
+  private Object convertToJsonNull(Object defaultValue) {
+    if (defaultValue instanceof Map) {
+      for (Map.Entry<Object, Object> entry : ((Map<Object, Object>) defaultValue).entrySet()) {
+        if (entry.getValue() instanceof Map || entry.getValue() instanceof Collection) {
+          entry.setValue(convertToJsonNull(entry.getValue()));
+        } else {
```
rzhang10, line 246: wouldn't you need a condition sth like `else if (entry.getValue() == null)`? same for line 258

rzhang10 (commented):
For literal `null` values (not complex objects), it will now be converted to `JsonProperties.NULL_VALUE`, have you tested if this causes regression or not?

A good way to verify this is to add 2 unit tests in `TestSparkAvroReaderForFieldsWithDefaultValue`. One that contains the primitive `null` default value, another that contains the case you are fixing, which is complex objects that contains `null` nested inside.

core/src/main/java/org/apache/iceberg/avro/TypeToSchema.java
zheliu2: This change should be in-place, so if there is no change needed, we don't have to check that branch.

zheliu2:
> if the table contains non-primitive Avro types (referred to as `TypeA`) as the first field and arrays of `TypeA` as the second field, it leads to a NullPointerException during loading the table schema.

Sure, that's a good catch

rzhang10 (approved):
Thanks for the fix.

PR #152, opened 2024-01-10
Add single union schema conversion logic for hive to iceberg

rzhang10:
This patch adds the parity conversion logic that already exists in avro/orc -> iceberg schema conversion where it treats the single type union as the unwrapped single type itself, instead of returning a struct of tag and field0.


zheliu2 (approved):
LGTM

ljfgem (approved):
LGTM

PR #153, opened 2024-03-11
[LI] Make MergeHiveSchemaWithAvro not reordering avro optional union …

rzhang10:
…field when it's it's merging a hive primitive with an avro optional union

This patch fixes a issue when a table's avro schema exhibit this pattern: it has a record field which has a inner optional field with null as the second element, and it has a default value at the record level with a key-value pair for that inner optional field, then the current `MergeHiveSchemaWithAvro` will reorder the null to be the first element of the optional field schema, making the record-level default validation fail. This patch will retain the positional of the `null` element in the avro optional field its original position, thus fixing the bug.


zheliu2 (approved):
LGTM

PR #154, opened 2024-03-20
[LI] Fix SPARK ORC projection read for deeply nested union schema and…

rzhang10:
… out-of-order projection fields

This patch fixes multiple issues currently in the Spark ORC union type projection read code #150 

### Issue 1:
Given an ORC schema of `struct<unionCol:uniontype<int,string>>`, the converted iceberg schema is to be `struct<0: tag: required int, 1: field0: optional int, 2: field1: optional string>` (field id not important here).
So, for a fully-supported projection feature, user should be arbitrarily permutate and partially project a subset of the [tag, field0, field1].
e.g. users can project:

- [tag, field0, field1] fully projected, original/in-order schema
- [tag] single field
- [tag, field0] 2 fields
- [field1, field0] 2 fields, but out of order
- [field0, field1, tag] full projection, but out of order/fields-permutation

The current code exhibits multiple bugs when projecting a subset fields (e.g. projecting field0 works but field1 fails), projecting out-of-order fields (e.g. [field1, field0]).

### Issue 2:
Given a deeply nested ORC schema with at least 2 levels of `uniontype`, e.g. `struct<c1:uniontype<int,struct<c2:string,c3:uniontype<int,string>>>>` the current code completely fails with NPE.

We should expect the user to do full/partial/out-of-order projections on 2/arbitrary levels of unions without any issue.

### Solution
This PR holistically fixes the above 2 issues by changing the `OrcSchemaWithTypeVisitor` and `SparkOrcValueReaders`, to make the first one always return a fully projected reader options to the reader code, so that the reader can determine from the expected schema, which specific reader's value to return (by juding the `tag` value), the `OrcSchemaWithTypeVisitor` implements its magic via a new visitor `OrcToIcebergVisitorWithPseudoId`, which can convert a orc schema with uniontype to iceberg, without preexisting `iceberg-id` metadata in the orc schema.

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
@@ -19,11 +19,14 @@
 
 package org.apache.iceberg.orc;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.types.Type;
 import org.apache.iceberg.types.Types;
+import org.apache.iceberg.types.Types.StructType;
 import org.apache.orc.TypeDescription;
 
 public abstract class OrcSchemaWithTypeVisitor<T> {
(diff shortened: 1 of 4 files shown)


orc/src/main/java/org/apache/iceberg/orc/OrcToIcebergVisitorWithPseudoId.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.orc.TypeDescription;
+
+public class OrcToIcebergVisitorWithPseudoId extends OrcToIcebergVisitor {
+
+  private static final String PSEUDO_ICEBERG_FIELD_ID = "-1";
+
+  @Override
+  public Optional<NestedField> union(TypeDescription union, List<Optional<NestedField>> options) {
+    union.setAttribute(org.apache.iceberg.orc.ORCSchemaUtil.ICEBERG_ID_ATTRIBUTE, PSEUDO_ICEBERG_FIELD_ID);
+    List<Optional<NestedField>> optionsCopy = new ArrayList<>(options);
```
1fanwang, line 37: nit:
more efficient to create an immutable copy of the list directly
```suggestion
    List<Optional<NestedField>> optionsCopy = List.copyOf(options);
```

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcValueReaders.java
```diff
@@ -169,38 +169,26 @@ protected void set(InternalRow struct, int pos, Object value) {
...
-    private boolean isTagFieldProjected;
-    private int numOfFieldsInReturnedRow;
+    private final Type expectedType;
+    private Map<Integer, Integer> idxInExpectedSchemaToIdxInReaders;
 
     private UnionReader(List<OrcValueReader<?>> readers, Type expected) {
       this.readers = new OrcValueReader[readers.size()];
       for (int i = 0; i < this.readers.length; i += 1) {
         this.readers[i] = readers.get(i);
       }
```
1fanwang: nit: simplify
```suggestion
      this.readers = readers.toArray(new OrcValueReader[0]);
```

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
```diff
@@ -525,6 +531,87 @@ public void testDeepNestedSingleTypeUnion() throws IOException {
     }
   }
 
+  @Test
```
1fanwang, line 534: Just for my learning:
this new test case seems to cover Issue 2 in the PR, do we also need additional unit test for testing single lvl union with partial projection for Issue 1?

1fanwang (commented):
Thanks for the PR! Minor comments non blocking

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
```diff
@@ -137,16 +138,15 @@ private <T> void visitComplexUnion(Type type, TypeDescription union, OrcSchemaWi
...
                                                               OrcSchemaWithTypeVisitor<T> visitor,
                                                               List<T> options,
                                                               int typeIndex) {
-    OrcToIcebergVisitor schemaConverter = new OrcToIcebergVisitor();
-    schemaConverter.beforeField("field" + typeIndex, schema);
-    schema.setAttribute(org.apache.iceberg.orc.ORCSchemaUtil.ICEBERG_ID_ATTRIBUTE, PSEUDO_ICEBERG_FIELD_ID);
-    Optional<Types.NestedField> icebergSchema = OrcToIcebergVisitor.visit(schema, schemaConverter);
-    schemaConverter.afterField("field" + typeIndex, schema);
-    options.add(visit(icebergSchema.get().type(), schema, visitor));
+    OrcToIcebergVisitor schemaConverter = new OrcToIcebergVisitorWithPseudoId();
```
zheliu2, line 145: can OrcToIcebergVisitorWithPseudoId be merged with OrcToIcebergVisitor? or do we create a new class for a reason?

spark/src/test/java/org/apache/iceberg/spark/data/TestSparkOrcUnions.java
rzhang10: I think there's previous unit test already testing the single lv union.

orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java
rzhang10: The `OrcToIcebergVisitor` is an existing native iceberg visitor. What we are doing here is linkedin specific (at least for now), to transform the orc type back to iceberg type which the orc type can contain unions. All these behavior and logic are not standard in native iceberg and we should use a separate linkedin specific subclass for now.

orc/src/main/java/org/apache/iceberg/orc/OrcToIcebergVisitorWithPseudoId.java
rzhang10: Are you saying the guava api? guava doesn't have this api i just check, it only has `Lists.newArrayList()` . which is the same as `new arraylist` i think.

rzhang10: and, in fact, the next line, i need to insert an element at the start of the List, so it cannot be immutable.

spark/src/main/java/org/apache/iceberg/spark/data/SparkOrcValueReaders.java
rzhang10: i can change it to
```
      this.readers = readers.toArray(new OrcValueReader[readers.size()]);
```

zheliu2 (approved):
LGTM

PR #156, opened 2024-12-02
Add logging for HiveMetadataPreservingTableOperations

rzhang10:
This PR adds comprehensive logging to the `HiveMetadataPreservingTableOperations` so that we get the timestamps of all the critical HMS client calls at their starting and end time, so that it can help us reason about what are the sequence of calls the actual code runtime does when running in the gobblin job, and their respective time taken.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ buildscript {
   }
   dependencies {
     classpath 'gradle.plugin.com.github.jengelman.gradle.plugins:shadow:7.0.0'
-    classpath 'com.palantir.baseline:gradle-baseline-java:4.0.0'
+    classpath 'com.palantir.baseline:gradle-baseline-java:4.40.0'
     classpath 'com.palantir.gradle.gitversion:gradle-git-version:0.12.3'
     classpath 'com.diffplug.spotless:spotless-plugin-gradle:6.8.0'
     classpath 'gradle.plugin.org.inferred:gradle-processors:3.3.0'
--- a/bundled-guava/src/main/java/org/apache/iceberg/GuavaClasses.java
+++ b/bundled-guava/src/main/java/org/apache/iceberg/GuavaClasses.java
@@ -55,6 +55,7 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 // inspired in part by
 // https://github.com/apache/avro/blob/release-1.8.2/lang/java/guava/src/main/java/org/apache/avro/GuavaClasses.java
+@SuppressWarnings("ReturnValueIgnored")
 public class GuavaClasses {
 
   /*
--- a/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
+++ b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
@@ -35,7 +35,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
-import jline.internal.Log;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.common.StatsSetupConst;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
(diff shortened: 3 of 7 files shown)


bundled-guava/src/main/java/org/apache/iceberg/GuavaClasses.java
```diff
@@ -55,6 +55,7 @@
 
 // inspired in part by
 // https://github.com/apache/avro/blob/release-1.8.2/lang/java/guava/src/main/java/org/apache/avro/GuavaClasses.java
+@SuppressWarnings("ReturnValueIgnored")
```
rzhang10, line 58: Need to add this else the build is failing, since we bumped the `gradle-baseline-java` plugin.

build.gradle
```diff
@@ -29,7 +29,7 @@ buildscript {
   }
   dependencies {
     classpath 'gradle.plugin.com.github.jengelman.gradle.plugins:shadow:7.0.0'
-    classpath 'com.palantir.baseline:gradle-baseline-java:4.0.0'
+    classpath 'com.palantir.baseline:gradle-baseline-java:4.40.0'
```
rzhang10, line 32: Need to bump this since `4.0.0` is no longer in maven at all.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
```diff
@@ -195,17 +195,36 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
     ReentrantLock tableLevelMutex = commitLockCache.get(fullName, t -> new ReentrantLock(true));
     tableLevelMutex.lock();
     try {
+      LOG.warn(
+          "In thread {}, starting to acquire a lock for table {}",
```
zheliu2, line 199: Nit: Is it possible to add timestamp before/after acquired the lock?

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
```diff
@@ -195,17 +195,36 @@ protected void doCommit(TableMetadata base, TableMetadata metadata) {
...
+          lockId.get(),
+          fullName);
       // TODO add lock heart beating for cases where default lock timeout is too low.
       Table tbl;
       // [LINKEDIN] Instead of checking if base != null to check for table existence, we query
       // metastore for existence
       // base can be null if not Iceberg metadata exists, but Hive table exists, so we want to get
       // the current table
       // definition and not create a new definition
+      LOG.warn("In thread {}, checking if table exists {}", Thread.currentThread(), fullName);
```
zheliu2, line 215: Nit: should we change this to level info, it seems like not strongly related to the locking issue

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
```diff
@@ -568,9 +568,20 @@ protected long acquireLock() throws UnknownHostException, TException, Interrupte
             Lists.newArrayList(lockComponent),
             System.getProperty("user.name"),
             InetAddress.getLocalHost().getHostName());
+    LOG.warn(
```
zheliu2, line 571: Nit: maybe add timestamp for lock method?

zheliu2 (approved):
LGTM, left a few nit comment

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
rzhang10: The timestamp should already be included as a prefix in the logging message?

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
rzhang10: I think let's consistency use `warn`, so that we can analyze the issue easier, also when we are done with finding out the RC, we should revert this PR.

hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java
rzhang10: same as above.

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
zheliu2: Ack, make sense

hivelink-core/src/main/java/org/apache/iceberg/hivelink/core/HiveMetadataPreservingTableOperations.java
zheliu2: Thanks, just checked the container log emitted by Gobblin, looks you are correct

zheliu2 (approved):

zheliu2 (approved):

zheliu2 (approved):

PR #189, opened 2025-09-23
Support budgeted file rewrite, ordering by file sequence number

teamurko:
This PR addresses the following use case that we have at LinkedIn.

A large unpartitioned table, 40k 500MB files, and hourly merge-into job that upserts/updates 95% of existing data files. Delete files created are position deletes and partition scoped (file scoped create too many files, but with file scoped deletes https://github.com/apache/iceberg/blob/042f01a240439691ac67c6d67bd51cf5a4167dd2/core/src/main/java/org/apache/iceberg/actions/BinPackRewriteFilePlanner.java#L270 allows to filter data files by number of deleted rows).

Each merge-into creates 100s of delete files. This is already many, if file open cost is 1MB (very optimistic), then it's equivalent of reading 100MB. Now it happens every time Spark executor scans a data file, it will have to open all 100s delete files in order to build delete index. Delete files pruning at plan time doesn't filter out many delete files due to how scattered keys are. We can't accumulate many delete files because they are associated with almost all data files, so we run minor compaction (position delete files rewrite), this helps. But then all data files have the same number of delete files associated, and filtering by https://github.com/apache/iceberg/blob/042f01a240439691ac67c6d67bd51cf5a4167dd2/core/src/main/java/org/apache/iceberg/actions/BinPackRewriteFilePlanner.java#L70 doesn't help.

We need to compact data files with the most number of deleted rows. The position delete files are partition scoped, so we can't look at deletes file stats for a specific data file. We use proxy for how many deleted rows are accumulated, that is how old data file is: file sequence number. We not only order file groups, we also sort tasks before grouping to maximize the effect of ordering.

On V3 Iceberg, we'll use DV with blobs from different data files in one DV file and use `delete-ratio-threshold`


api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
```diff
@@ -70,6 +70,11 @@ public interface RewriteDataFiles
 
   long MAX_FILE_GROUP_SIZE_BYTES_DEFAULT = 1024L * 1024L * 1024L * 100L; // 100 Gigabytes
 
+  /** The maximum total size of all file groups that this rewrite is allowed to process. */
+  String MAX_TOTAL_FILE_GROUP_SIZE_BYTES = "max-total-file-group-size-bytes";
+
+  long MAX_TOTAL_FILE_GROUP_SIZE_BYTES_DEFAULT = 0L; // no limit
```
sumedhsakdeo: MAX_LONG?

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
```diff
@@ -237,6 +242,9 @@ interface FileGroupInfo {
     /** returns which file group this is out of the set of file groups for this partition */
     int partitionIndex();
 
+    /** returns the minimum data sequence number of the files in this file group */
```
sumedhsakdeo: We saw there is a data sequence number and file sequence number? Can we use exact terminology here and other place?
- Data sequence: Represents when the data was logically added
- File sequence: Represents when the file was physically added

I believe we want data sequence number.

sumedhsakdeo: https://github.com/linkedin/iceberg/pull/189/files#diff-0875d7e14b49fc353c79bae33cc9fc5be974c42a66b031dec54465a5fd9ca347L108 needs an update.

core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewriter.java
```diff
@@ -159,8 +162,18 @@ protected boolean wronglySized(T task) {
   @Override
   public Iterable<List<T>> planFileGroups(Iterable<T> tasks) {
     Iterable<T> filteredTasks = rewriteAll ? tasks : filterFiles(tasks);
+    // TODO: pass use comparator consistent with FG comparator
```
sumedhsakdeo: One option is to instead of global ordering based on sequence number, we do ordering of pre-computed rewritefilegroups.

https://github.com/linkedin/iceberg/blob/cbb853073e681b4075d7c8707610dceecbee3a82/core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java#L98

If we want global ordering,  can we create a SequenceAwareFileRewritter extend it from SizeBasedFileRewriter, and implement the sorting logic in overridden planFileGroups?

sumedhsakdeo: Should SizeBasedFileRewriter class have a member totalFileGroupBytesToRewrite, and handle that option https://github.com/linkedin/iceberg/pull/189/files#diff-60045ba49ea7ae9fb5232ecace068e9deac5a697aca7582eca92106cab28e8d2R132?

sumedhsakdeo (changes requested):
Thanks Stas, left some early thoughts. Wondering why this PR cannot be done in upstream apache/iceberg and then backported? Is there a significant refactor in the upstream?

teamurko:
> Thanks Stas, left some early thoughts. Wondering why this PR cannot be done in upstream apache/iceberg and then backported? Is there a significant refactor in the upstream?

there is a risk that upstream PR will take much longer time to review

sumedhsakdeo:
> > Thanks Stas, left some early thoughts. Wondering why this PR cannot be done in upstream apache/iceberg and then backported? Is there a significant refactor in the upstream?
> 
> 
> 
> there is a risk that upstream PR will take much longer time to review

Ok, but we can always bring it in here if we think it is ready. I think there is value in starting this one upstream.

cbb330:
the 1.10.0 PR: https://github.com/apache/iceberg/pull/12824

is count based limit, while this PR is size based.

the count based approach is able to limit in pre-planning, which is more efficient in IOPs. do we have size metrics in pre-planning stage to do the same optimization, so the logic will be filtered starting in BinPackRewriteFilePlanner?

also, is there a benefit to limit number of bytes (but allowing any number of file groups)?

teamurko:
> the 1.10.0 PR: [apache/iceberg#12824](https://github.com/apache/iceberg/pull/12824)
> 
> is count based limit, while this PR is size based.
> 
> the count based approach is able to limit in pre-planning, which is more efficient in IOPs. do we have size metrics in pre-planning stage to do the same optimization, so the logic will be filtered starting in BinPackRewriteFilePlanner?
> 
> also, is there a benefit to limit number of bytes (but allowing any number of file groups)?

file size is part of metadata, it's available at planning time, limiting by count will yield uneven workloads when files have different sizes, e.g. append files as part of merge-in, they are pretty small.

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
```diff
@@ -172,6 +175,23 @@ public RewriteDataFiles.Result execute() {
 
     Stream<RewriteFileGroup> groupStream = toGroupStream(ctx, fileGroupsByPartition);
 
+    // if maxTotalFileGroupSizeBytes is set, limit the total size of the file groups
+    if (maxTotalFileGroupSizeBytes > 0) {
```
cbb330: current strategy is:
1. collect file group stream
2. trim group stream

downside to this is: 
1. the smallest unit to filter on is a whole filegroup. a file group can be very large (default is 100gb for single max size)

so assume that a user configures rewrite_data_files in such a way that the file groups are large, then attempts to use the filter for total file group size, they could end up with 0 work. 

I wonder if it is strictly beneficial to do instead:
1. trim planned files
2. collect file group stream

in this model, the file group stream can still be dynamically sized (large file groups, small file groups, etc.) but the amortizable metric of file size is still utilized

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
```diff
@@ -105,6 +109,26 @@ public static Comparator<RewriteFileGroup> comparator(RewriteJobOrder rewriteJob
...
+  }
+
+  public static Comparator<FileScanTask> taskComparator(RewriteJobOrder rewriteJobOrder) {
+    switch (rewriteJobOrder) {
+      case FILES_ASC:
+        return Comparator.comparing(FileScanTask::length);
+      case FILES_DESC:
+        return Comparator.comparing(FileScanTask::length, Comparator.reverseOrder());
+      case FILES_MIN_SEQUENCE_NUMBER_ASC:
+        return Comparator.comparing(t -> t.file().dataSequenceNumber());
```
sumedhsakdeo: Should this be fileSequenceNumber?

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
```diff
@@ -105,6 +109,26 @@ public static Comparator<RewriteFileGroup> comparator(RewriteJobOrder rewriteJob
...
+        return (fileGroupOne, fileGroupTwo) -> 0;
+    }
+  }
+
+  public static Comparator<FileScanTask> taskComparator(RewriteJobOrder rewriteJobOrder) {
+    switch (rewriteJobOrder) {
+      case FILES_ASC:
+        return Comparator.comparing(FileScanTask::length);
+      case FILES_DESC:
+        return Comparator.comparing(FileScanTask::length, Comparator.reverseOrder());
```
sumedhsakdeo:
```suggestion
      case BYTES_ASC:
        return Comparator.comparing(FileScanTask::length);
      case BYTES_DESC:
        return Comparator.comparing(FileScanTask::length, Comparator.reverseOrder());
```

FileScanTask::length matches expectation for BYTES_ASC and BYTES_DESC.

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
```diff
@@ -227,7 +247,19 @@ private StructLikeMap<List<FileScanTask>> groupByPartition(
...
   private StructLikeMap<List<List<FileScanTask>>> fileGroupsByPartition(
       StructLikeMap<List<FileScanTask>> filesByPartition) {
-    return filesByPartition.transformValues(this::planFileGroups);
+    // Keep ordering of files in line with file group ordering
+    // 1. If file groups are ordered by number of files, then the goal is to rewrite smaller
+    // (larger) files first,
+    //   in this case order files by size
+    // 2. If file groups are ordered by min sequence number, the goal is to rewrite older (newer)
+    // files first,
+    //   in this case order files by data sequence number
```
sumedhsakdeo:
```suggestion
    // Align file ordering with the chosen file group ordering:
    // 1. If groups are ordered by bytes, rewrite smaller (or larger) files first → sort by file size.
    // 2. If groups are ordered by minimum sequence number, rewrite older (or newer) files first → sort by sequence number.
```

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
```diff
@@ -227,7 +247,19 @@ private StructLikeMap<List<FileScanTask>> groupByPartition(
...
+    return filesByPartition.transformValues(
+        tasks ->
+            this.planFileGroups(
+                tasks.stream()
+                    .sorted(RewriteFileGroup.taskComparator(rewriteJobOrder))
```
sumedhsakdeo: I’m concerned about the memory overhead here — a global sort across tens of millions of FileScanTasks could lead to OOMs, and increase in the planning time.

Instead of fully sorting, we could:

Push the ordering logic down into planFileGroups, applying only the coarser ordering required by rewriteJobOrder (e.g., by minDataSequenceNumber like in this PR or use the population of FileScanTasks within each RewriteFileGroup to rank groups relative to one another).

In practice, an approximate ordering should be sufficient.

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
teamurko: I was not sure about the purpose of bytes ordering, rewriting smaller/larger file groups first? With that it's not clear how tasks sorting helps. For file count ordering I can see that it might be beneficial to rewrite smaller files first, or break down larger files first, and sorting by task length helps.

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
teamurko: good point.. data seq number is used for deletes to determine if delete applies, file seq number is used for versioning, i guess file seq number would be more appropriate here

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
teamurko: I thought about partial ordering. This sort applies to task for a single partition, in practice single partition doesn't have MMs of files, our largest CDC tables have < 100K files total. All tasks are materialized into a list already before sort is called. Sort can use some stack, but doesn't create copies (except maybe references, but it's still < 1M records), and it's fast.

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
teamurko: Thanks, @cbb330 

1) filter tasks before grouping by partition
 - pros
   - more accurate, allows budget < file group size
 - cons
   - tasks can be fragmented => uneven file groups sizes

For use case of data rewrite to merge small files, the fragmentation is harmful.
We can require budget >= file group size, or truncate groups.

core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewriter.java
teamurko: the comparator for tasks is moved to the file group class close to the fg comparator

api/src/main/java/org/apache/iceberg/actions/RewriteDataFiles.java
teamurko: switched to file sequence number, mentioned in a separate thread

core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewriter.java
teamurko: i tried to make it a little more consistent b/w tasks and fg ordering so that we don't need to have another implementation and choose new one based on user provided options, i think sorting tasks for files-asc/desc fg ordering is beneficial

core/src/main/java/org/apache/iceberg/actions/RewriteFileGroup.java
```diff
@@ -105,6 +109,26 @@ public static Comparator<RewriteFileGroup> comparator(RewriteJobOrder rewriteJob
...
```
sumedhsakdeo:
```suggestion
```

core/src/main/java/org/apache/iceberg/actions/RewritePositionDeletesGroup.java
```diff
@@ -126,6 +130,26 @@ public static Comparator<RewritePositionDeletesGroup> comparator(RewriteJobOrder
...
+  public static Comparator<PositionDeletesScanTask> taskComparator(RewriteJobOrder order) {
+    switch (order) {
+      case FILES_ASC:
+        return Comparator.comparing(t -> t.file().fileSizeInBytes());
+      case FILES_DESC:
+        return Comparator.comparing(t -> t.file().fileSizeInBytes(), Comparator.reverseOrder());
+      case FILES_MIN_SEQUENCE_NUMBER_ASC:
+        return Comparator.comparing(t -> t.file().fileSequenceNumber());
+      case FILES_MIN_SEQUENCE_NUMBER_DESC:
+        return Comparator.comparing(t -> t.file().fileSequenceNumber(), Comparator.reverseOrder());
```
sumedhsakdeo:
```suggestion
```

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
teamurko: Discussed offline, we'll go more granular, limit tasks, then do grouping

cbb330 (approved):
LGTM!

core/src/main/java/org/apache/iceberg/actions/RewritePositionDeletesGroup.java
teamurko: Updated

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
teamurko: it's now removed, the sorting happens before grouping

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
```diff
@@ -188,10 +190,29 @@ StructLikeMap<List<List<FileScanTask>>> planFileGroups(long startingSnapshotId)
...
+    long pickedTasksSizeBytes = 0;
+    int pickedTasksCount = 0;
+    for (FileScanTask task : tasks) {
+      if (pickedTasksSizeBytes + task.sizeBytes() <= maxTotalFilesSizeBytes) {
+        pickedTasksCount++;
+        pickedTasksSizeBytes += task.sizeBytes();
+      } else {
+        break;
+      }
+    }
```
sumedhsakdeo, line 209:
```suggestion
      if ( maxTotalFilesSizeBytes < MAX_TOTAL_FILES_SIZE_BYTES_DEFAULT ) { 
          long pickedTasksSizeBytes = 0;
          int pickedTasksCount = 0;
          for (FileScanTask task : tasks) {
               if (pickedTasksSizeBytes + task.sizeBytes() <= maxTotalFilesSizeBytes) {
                   pickedTasksCount++;
                   pickedTasksSizeBytes += task.sizeBytes();
               } else {
                   break;
               }
             }
             tasks = tasks.subList(0, pickedTasksCount);
        }     
```

sumedhsakdeo (approved):
one minor comment otherwise looks good. thanks @teamurko

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/RewriteDataFilesSparkAction.java
teamurko: This would couple this code with MAX_TOTAL_FILES_SIZE_BYTES_DEFAULT value being max long.

sumedhsakdeo:
Do you plan to backport this in 1.2?

teamurko:
> Do you plan to backport this in 1.2?

The app will run on 1.5.2+Spark3.5, if it's not the case, I can port to 1.2

PR #191, opened 2025-10-02
Emit metrics of uncompressed bytes written for spark writes

jiang95-dev:
## Description
There can be many ways to optimize the file size when doing spark write. For example, using particular file format with some codec, or applying bucketing/sorting strategies. But we need to know the base metrics for evaluation purposes. So we start to emit metrics for uncompressed bytes written.

This approach sums the bytes for each `InternalRow`. This data structure is highly optimized, whereas ORC files can contain metadata. So it is very likely the compression rate can be seemingly negative, but it is ok since we are comparing across different techniques.

The `totalUncompressedBytesWritten` will be added as part of `CommitReport` and be emitted with the existing client-side audit event.

## Testing
Added unit tests. Manually checked the `CommitReport` in `LoggingMetricsReporter` and saw the `totalUncompressedBytesWritten` field exists.


core/src/main/java/org/apache/iceberg/metrics/CommitMetricsResult.java
```diff
@@ -50,6 +50,7 @@ public interface CommitMetricsResult {
   String ADDED_EQ_DELETES = "added-equality-deletes";
   String REMOVED_EQ_DELETES = "removed-equality-deletes";
   String TOTAL_EQ_DELETES = "total-equality-deletes";
+  String TOTAL_UNCOMPRESSED_BYTES_WRITTEN = "total-uncompressed-bytes-written";
```
teamurko:
```suggestion
  String ADDED_UNCOMPRESSED_BYTES = "added-uncompressed-bytes";
```

teamurko: It's only for added data files, afaiu, right?

core/src/main/java/org/apache/iceberg/SnapshotSummary.java
```diff
@@ -58,6 +58,7 @@ public class SnapshotSummary {
   public static final String SOURCE_SNAPSHOT_ID_PROP = "source-snapshot-id";
   public static final String REPLACE_PARTITIONS_PROP = "replace-partitions";
   public static final String EXTRA_METADATA_PREFIX = "snapshot-property.";
+  public static final String TOTAL_UNCOMPRESSED_BYTES_WRITTEN = "total-uncompressed-bytes-written";
```
teamurko:
```suggestion
  public static final String ADDED_UNCOMPRESSED_BYTES_PROP = "added-uncompressed-bytes";
```

teamurko: ditto

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -166,7 +168,8 @@ private WriterFactory createWriterFactory() {
         tableBroadcast, format, targetFileSize, writeSchema, dsSchema, partitionedFanoutEnabled);
   }
 
-  private void commitOperation(SnapshotUpdate<?> operation, String description) {
+  private void commitOperation(
+      SnapshotUpdate<?> operation, String description, long totalUncompressedBytesWritten) {
```
teamurko: SnapshotUpdate and it's subclasses allow setting a property, I'm thinking that it'd be cleaner to set the prop higher in the call stack, so that these methods don't need to change, there should be much less code change amplification

teamurko (changes requested):
This is great, thank you for the quick PR. Left a few comments.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -696,6 +737,7 @@ private PartitionedDataWriter(
     @Override
     public void write(InternalRow row) throws IOException {
       partitionKey.partition(internalRowWrapper.wrap(row));
+      uncompressedBytesWritten += ((UnsafeRow) row).getSizeInBytes();
```
sumedhsakdeo, line 740: I am not sure if we can rely on deterministically to figure out uncompressed bytes from a Spark internal data structure. It also leave a gap for how we will do it for Trino. 
CC: @shardulm94

sumedhsakdeo (changes requested):
Not sure if this is the direction we want to pursue for tracking uncompressed bytes.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
rdsr: ORC has getRawDataSize method which gives the right uncompressed bytes after encoding but before compression. We can probably get this right after the ORC closes it file after writing. This would be at a much lower level in Iceberg [orc file writer?] but workable in both Spark and Trino and other engines/frameworks too which call into Iceberg. In that sense I wonder if the change is better suited at the Iceberg data file writer level...

thoughts?

teamurko: This is interesting, we can propagate raw size in metrics https://github.com/linkedin/iceberg/blob/cbb853073e681b4075d7c8707610dceecbee3a82/orc/src/main/java/org/apache/iceberg/orc/OrcFileAppender.java#L110
Good option that works for all supported engines

If we sort rows, and say dicts become smaller, then I guess it will reduce raw size, right?

jiang95-dev:
Closing the PR since we are not on the right track. The uncompressed-bytes should be engine-agnostic.

PR #219, opened 2025-12-01
Add support to configure replication factor of delete files.

shanthoosh:
**Changes**

Adds support to configures the replication factor during the creation of the delete files. Enables the spark jobs to set the HDFS delete replication as part of the writer configuration.

**Tests**
- Added the unit tests for all the introduced changes. 
- Verified that existing FileIO implementations continue to work.
- ./gradlew clean && ./gradlew build


api/src/main/java/org/apache/iceberg/io/FileIO.java
```diff
@@ -73,6 +73,18 @@ default InputFile newInputFile(ManifestFile manifest) {
...
+   * Get a {@link OutputFile} instance to write bytes to the file at the given path with specific
+   * properties.
+   *
+   * @param path the file path
+   * @param properties the properties for the output file
+   * @return an OutputFile for writing data
+   */
+  default OutputFile newOutputFile(String path, Map<String, String> properties) {
+    return newOutputFile(path);
+  }
```
sumedhsakdeo, line 89: shouldn't the default impl throw exception?
or the other way around which might be a much larger change

```
OutputFile newOutputFile(String path) {
    return newOutputFile(path, new HashMap<>());
} 
```

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
```diff
@@ -54,24 +58,53 @@ public static OutputFile fromPath(Path path, Configuration conf) {
     return fromPath(path, fs, conf);
   }
 
+  public static OutputFile fromPath(Path path, Configuration conf, Map<String, String> properties) {
+    short replicationFactor = -1;
```
sumedhsakdeo: This should not be -1, I think we should set it to either default replication factor from conf, if it is not available in conf, have a constant for DEFAULT_REPLICATION_FACTOR

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
```diff
@@ -54,24 +58,53 @@ public static OutputFile fromPath(Path path, Configuration conf) {
...
   public PositionOutputStream create() {
     try {
-      return HadoopStreams.wrap(fs.create(path, false /* createOrOverwrite */));
+      if (replication > 0) {
+        return HadoopStreams.wrap(
+            fs.create(
+                path,
+                false /* overwrite */,
+                conf.getInt(
+                    CommonConfigurationKeysPublic.IO_FILE_BUFFER_SIZE_KEY,
```
sumedhsakdeo, line 113: why was this added?

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
```diff
@@ -54,24 +58,53 @@ public static OutputFile fromPath(Path path, Configuration conf) {
...
+                    CommonConfigurationKeysPublic.IO_FILE_BUFFER_SIZE_DEFAULT),
+                replication,
+                fs.getDefaultBlockSize(path)));
```
sumedhsakdeo, line 116: ditto

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
```diff
@@ -82,7 +115,19 @@ public PositionOutputStream create() {
...
   public PositionOutputStream createOrOverwrite() {
     try {
-      return HadoopStreams.wrap(fs.create(path, true /* createOrOverwrite */));
+      if (replication > 0) {
+        return HadoopStreams.wrap(
+            fs.create(
+                path,
+                true /* overwrite */,
+                conf.getInt(
+                    CommonConfigurationKeysPublic.IO_FILE_BUFFER_SIZE_KEY,
```
sumedhsakdeo, line 136: ditto

core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java
```diff
@@ -73,7 +79,8 @@ private OutputFileFactory(
       int partitionId,
       long taskId,
       String operationId,
-      String suffix) {
+      String suffix,
+      Optional<Short> replicationFactorOptional) {
```
sumedhsakdeo: nit just call it replicationFactor

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java
```diff
@@ -102,6 +102,8 @@ class SparkPositionDeltaWrite implements DeltaWrite, RequiresDistributionAndOrde
   private final Map<String, String> extraSnapshotMetadata;
   private final Distribution requiredDistribution;
   private final SortOrder[] requiredOrdering;
+  private final Map<String, String> writeProperties;
```
sumedhsakdeo: why?

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
```diff
@@ -66,4 +66,16 @@ private SparkSQLProperties() {}
 
   // Controls whether to report locality information to Spark while allocating input partitions
   public static final String LOCALITY = "spark.sql.iceberg.locality.enabled";
+
+  // Controls compression codec for write operations
+  public static final String COMPRESSION_CODEC = "spark.sql.iceberg.compression-codec";
```
sumedhsakdeo: this should not be part of this PR.

sumedhsakdeo (changes requested):
Left a few comments.

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
teamurko: wonder where it connects with ORC's block size

api/src/main/java/org/apache/iceberg/io/FileIO.java
shanthoosh: Fixed.

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
shanthoosh: Fixed

core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java
shanthoosh: Fixed.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeltaWrite.java
shanthoosh: Without these cherry-pick changes, building the spark-3.3 module was failing. Let me remove the spark-3.3 changes. I will submit a separate PR for these changes.

spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkSQLProperties.java
shanthoosh: Without these cherry-pick changes, building the spark-3.3 module was failing. Sure, let me remove the spark-3.3 changes. I will submit a separate PR for these changes.

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
shanthoosh: Added the change to wire-in the replication-factor argument to create file API. The additional additional params were default values used for the params. 

```
create(f, overwrite, 
                  getConf().getInt("io.file.buffer.size", 4096),
                  getDefaultReplication(f),
                  getDefaultBlockSize(f)
```

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
shanthoosh: Added the change to wire-in the replication-factor argument to create file API. The additional additional params were default values used for the params.
```java
create(f, overwrite, 
                  getConf().getInt("io.file.buffer.size", 4096),
                  getDefaultReplication(f),
                  getDefaultBlockSize(f)

```

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
```diff
@@ -54,24 +59,53 @@ public static OutputFile fromPath(Path path, Configuration conf) {
     return fromPath(path, fs, conf);
   }
 
+  public static OutputFile fromPath(Path path, Configuration conf, Map<String, String> properties) {
+    short replicationFactor = DEFAULT_REPLICATION_FACTOR;
+    if (properties != null) {
+      String replicationFactorAsString = properties.get(OutputFileFactory.FILE_REPLICATION_FACTOR);
+      if (replicationFactorAsString != null) {
+        replicationFactor = Short.parseShort(replicationFactorAsString);
```
sumedhsakdeo: Add a test-case for is parsing a short fails?

core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java
```diff
@@ -73,7 +79,8 @@ private OutputFileFactory(
       int partitionId,
       long taskId,
       String operationId,
-      String suffix) {
+      String suffix,
+      Optional<Short> replicationFactor) {
```
sumedhsakdeo, line 83: Why did we not pass the properties map to this layer, and instead passed replicationFactor?
Then, below, we can use isEmpty on the map to decide which newOutputFile method to call.

spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeletesRewrite.java
```diff
@@ -78,6 +78,7 @@ public class SparkPositionDeletesRewrite implements Write {
   private final int specId;
   private final StructLike partition;
   private final Map<String, String> writeProperties;
+  private final short replicationFactor;
```
sumedhsakdeo, line 81: Any reason why replicationFactor is not part of writeProperties [above](https://github.com/linkedin/iceberg/pull/219/changes#diff-fc0136e05fcd728d314e741a69e759af8d8ae126cf5dcc98f4a5ed8d0f14897fR80)?
what is the difference, what goes in writeProperties v/s what goes separate?

shanthoosh: Good question. The  writeProperties for both delete and data files, only contains configurations for write-format(avro, parquet, orc) and compression-related configuration such as compression-strategy, compression-level etc for different compression types(avro, parquet, orc). Relevant [control-flow
](https://github.com/linkedin/iceberg/blob/openhouse-1.5.2/spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L437C1-L529C1)
The other output-file parameters like target-file-size, delete-file-format, file-id etc were kept external to writeProperties as high level fields in the SparkPositionDeleteRewriter and in other spark writers. This is the reason why we kept replicationFactor separate and outside of writeProperties to be consistent with other write parameters. 

Also replication factor is a primitive of the file-creation lifecycle (I/O layer) rather than the data-persistence phase (serialization layer), it is excluded from the writeProperties map.

Please note this pattern applies to all spark writer classes in different spark versions 3.4, 3.5 and 3.3.

core/src/main/java/org/apache/iceberg/hadoop/HadoopOutputFile.java
shanthoosh: Sure, will do.

core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java
shanthoosh: The OutputFileFactory has associated builder(Here is the [relevant-control-flow](https://github.com/linkedin/iceberg/blob/openhouse-1.5.2/core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java#L121)) which is used by both spark and Flink writer classes to create both delete and data OutputFileFactories. The different file configurations such as fileFormat, suffix, encryption, location  are kept as separate high level fields and are not part of single writerPropertiesMap. Since replicationFactor is also an separate file-config, we added a separate setter method in builder and added it as high level field.

sumedhsakdeo (approved):
lgtm

spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/source/SparkPositionDeletesRewrite.java
shanthoosh: replication factor is used during file creation and not during writing data. WriteProperties only has file format (Avro, Parquet, ORC) and codec parameters used during writing phase. Here is the relevant control-flow as to where write-properties is used
```java
      OutputFileFactory deleteFileFactory =
          OutputFileFactory.builderFor(table, partitionId, taskId)
              .format(format)
              .operationId(queryId)
              .replicationFactor(replicationFactor)
              .suffix("deletes")
              .build();


      SparkFileWriterFactory writerFactoryWithRow =
          SparkFileWriterFactory.builderFor(table)
              .deleteFileFormat(format)
              .positionDeleteRowSchema(positionDeleteRowSchema)
              .positionDeleteSparkType(deleteSparkType)
              .writeProperties(writeProperties)
              .build();
      SparkFileWriterFactory writerFactoryWithoutRow =
          SparkFileWriterFactory.builderFor(table)
              .deleteFileFormat(format)
              .positionDeleteSparkType(deleteSparkTypeWithoutRow)
              .writeProperties(writeProperties)
              .build();
```

teamurko (approved):
Thanks @shanthoosh.

PR #221, opened 2025-12-11
Fix  to read retry properties from  metadata instead of , allowing tr…

sumedhsakdeo:
…ansaction-scoped retry configuration.

# Problem 
Currently, `commitSimpleTransaction()` ignores properties set via `transaction.updateProperties()` because it reads from `base` metadata. This prevents service-layer operations from using fail-fast (retry=0) strategy.

# Solution

Read retry config from `current` metadata in commit methods, respecting transaction-level property updates.
- Enables service/client retry separation


cbb330:
LGTM!

cbb330:
actually, can you apply this to the 1.5.2 branch?

OpenHouse is using 1.5.2 in tables-service

cbb330 (approved):

PR #223, opened 2025-12-12
Partially shade iceberg-hive-metastore for HiveSchemaUtils usage

jiang95-dev:
## Changes
There are some usages of `HiveSchemaUtils` in the hive to iceberg migration, however the `iceberg-hive-metastore` module did not properly include the dependencies in RuntimeClasspath. This PR adds shading to the module and relocate hive related classes so it won't break the existing classes in the hive environment. Also since `HiveSchemaUtils` is the only use case, we should only partially shade the jars so that we don't need to deal with massive exclusions and relocations.

## Tests
**Integration tests**
Tested the snapshot shaded jar in `li-openhouse-java-runtime` with `HiveSchemaUtils` tests and worked.

**Shadow jar check**
The necessary classes exist and being relocated. The jar size is 18mb compared to 36mb for full shading.
Screenshot 2025-12-12 at 1 59 41 AM

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -607,6 +607,18 @@ project(':iceberg-gcp') {
 }
 
 project(':iceberg-hive-metastore') {
+  apply plugin: 'com.github.johnrengelman.shadow'
+
+  tasks.shadowJar.dependsOn project(':iceberg-api').tasks.jar
+
+  configurations {
+    fatJarPackagedDependencies {}
+  }
+
+  ext {
+    hiveVersion = '2.3.8'
+  }
+
   dependencies {
     implementation project(path: ':iceberg-bundled-guava', configuration: 'shadow')
     implementation project(':iceberg-core')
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -649,28 +663,46 @@ project(':iceberg-hive-metastore') {
...
 
     testImplementation project(path: ':iceberg-api', configuration: 'testArtifacts')
+
+    fatJarPackagedDependencies("org.apache.hive:hive-metastore:" + shadedHiveVersion, commonExcludes)
+    fatJarPackagedDependencies(project(path: ':iceberg-bundled-guava', configuration: 'shadow'))
+    fatJarPackagedDependencies(project(':iceberg-core'))
+    fatJarPackagedDependencies(project(':iceberg-api'))
+    fatJarPackagedDependencies(project(':iceberg-common'))
+    fatJarPackagedDependencies("com.github.ben-manes.caffeine:caffeine")
+    fatJarPackagedDependencies("org.apache.hive:hive-storage-api:2.4.0")
```
teamurko: why is it not transitively pulled?

teamurko (approved):
if possible, we can have current module unchanges and produce shaded version in addition?

teamurko (approved):

PR #226, opened 2025-12-17
Fix bug in hive shading and make fat jar thinner

jiang95-dev:
## Changes
Last PR (https://github.com/linkedin/iceberg/pull/223) broke the build of `dali-data-sdk`. This PR excludes relocating the classes in `hive-metastore`, and make the fat jar thinner by only packaging `hive-serde` and `hive-common`.

## Tests
Tested build the module dali-data-sdk:sdk-dali-client.

PR #227, opened 2025-12-19
Hive shading continued

jiang95-dev:
## Changes
Follow up PR of https://github.com/linkedin/iceberg/pull/226. 
- Remove minimize() to bring in other serde classes.
- Add hive-storage-api to bring in TimestampUtils.
- Update relocation to only relocate hive related classes.

## Tests
Tested spark and unshaded spark with the snapshot jar. Called getTable for both timestamp and date-partitioned OH tables and hive tables.


aastha25 (approved):
tested internally

PR #228, opened 2026-01-08
Adding SparkSQLProperty for split-size

shanthoosh:
### Description

In Hive, we relied on the session-level config spark.sql.files.maxPartitionBytes to control file split sizes, and this was widely adopted across many flows at LinkedIn. As we transition to Iceberg, the equivalent setting is read.split.target-size, but it can't be set via session configs. This creates a gap—our current options are either to update thousands of jobs to pass this value through OPTIONS, or to set the table property spark.sql.iceberg.split-size directly on the tables and apply the config to all the jobs. Both of which is suboptimal.

Back-port PR: https://github.com/apache/iceberg/pull/13677
Same changes done in spark-3.1 for openhouse-1.2.0 branch https://github.com/linkedin/iceberg/pull/190 

### Testing
./gradlew clean && ./gradlew build


abhisheknath2011 (approved):
Thanks for the quick PR.

teamurko (approved):

PR #230, opened 2026-01-20
Add split size  session config support in SparkScanBuilder.configureSplitPlanning()

shanthoosh:
### Problem
The session config spark.sql.iceberg.split-size was honored in some Spark read paths but ignored in others. Specifically, SparkReadConf.splitSizeOption() only checked the read option (SparkReadOptions.SPLIT_SIZE) and not the session config, causing inconsistent behavior:  Specifically, APIs such as SparkStagedScan and SparkMicroBatchStream uses the session config(SparkSQLProperties.SPLIT_SIZE), while SparkScanBuilder.configureSplitPlanning() did not respect the session configuration(SparkSQLProperties.SPLIT_SIZE).

### Fix
This PR fixes the inconsistency by updating splitSizeOption() to also consider the session configuration (SparkSQLProperties.SPLIT_SIZE), ensuring consistent split-size handling across all Spark reader control flows.

### Testing

1. Added unit tests
2. Verified with ./gradlew clean && ./gradlew build

### DIFF
--- a/spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/SparkReadConf.java
+++ b/spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/SparkReadConf.java
@@ -194,7 +194,13 @@ public class SparkReadConf {
   }
 
   public Long splitSizeOption() {
-    return confParser.longConf().option(SparkReadOptions.SPLIT_SIZE).parseOptional();
+    return confParser
+        .longConf()
+        .option(SparkReadOptions.SPLIT_SIZE)
+        .sessionConf(SparkSQLProperties.SPLIT_SIZE)
+        .tableProperty(TableProperties.SPLIT_SIZE)
+        .defaultValue(TableProperties.SPLIT_SIZE_DEFAULT)
+        .parseOptional();
   }
 
   public long splitSize() {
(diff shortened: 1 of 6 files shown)


spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/SparkReadConf.java
```diff
@@ -194,7 +194,13 @@ public int orcBatchSize() {
   }
 
   public Long splitSizeOption() {
```
sumedhsakdeo, line 196: The name does mean that it is reading from Option alone.

shanthoosh: In SparkStagedScan and SparkMicroBatchStream, readConf.splitSize() method is used. Here is the relevant usage:

-  https://github.com/linkedin/iceberg/blob/openhouse-1.5.2/spark/v3.4/spark/src/main/java/org/apache/iceberg/spark/source/SparkStagedScan.java#L46  
- https://github.com/linkedin/iceberg/blob/openhouse-1.5.2/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/source/SparkMicroBatchStream.java#L103

sumedhsakdeo (changes requested):
We should take a deeper look at why only options / tableproperty are being looked at in the alternate split size path.

overriding method that is named to look at spark options alone with session conf / tblproperty, etc. seems incorrect.

PR #232, opened 2026-03-04
Fix RCE vulnerability in JMH Benchmarks workflow

vikrambohra:
## Summary
- Fix Remote Code Execution (RCE) vulnerability in the JMH Benchmarks GitHub Actions workflow
- Move all user-controlled `workflow_dispatch` inputs from direct `${{ }}` shell interpolation to `env:` variables, which GitHub automatically sanitizes against command injection
- Quote gradlew arguments to prevent word splitting

User-controlled inputs (`spark_version`, `benchmarks`, `repo`, `ref`) were directly interpolated into shell `run` commands, allowing arbitrary command injection (e.g., `iceberg-spark-3.3| id; cat /etc/passwd;#`).

Fixes: BDP-72545, PWN-46510

## Testing Done
- [ ] Verify workflow still triggers correctly via workflow_dispatch
- [ ] Confirm env vars are properly passed to shell commands
- [ ] Validate that malicious input in spark_version field no longer executes

### DIFF
--- a/.github/workflows/jmh-benchmarks.yml
+++ b/.github/workflows/jmh-benchmarks.yml
@@ -47,8 +47,10 @@ jobs:
           repository: ${{ github.event.inputs.repo }}
           ref: ${{ github.event.inputs.ref }}
       - id: set-matrix
+        env:
+          BENCHMARKS_INPUT: ${{ github.event.inputs.benchmarks }}
         run: |
-          matrix=$(echo '[${{ github.event.inputs.benchmarks }}]' | jq '.[] | select(endswith("Benchmark")) | .')
+          matrix=$(echo "[$BENCHMARKS_INPUT]" | jq '.[] | select(endswith("Benchmark")) | .')
           matrix=$(echo $matrix | sed 's/ /,/g' | sed 's/"/\"/g')
           echo "::set-output name=matrix::[$matrix]"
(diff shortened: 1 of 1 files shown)


.github/workflows/jmh-benchmarks.yml
```diff
@@ -47,8 +47,10 @@ jobs:
           repository: ${{ github.event.inputs.repo }}
           ref: ${{ github.event.inputs.ref }}
       - id: set-matrix
+        env:
+          BENCHMARKS_INPUT: ${{ github.event.inputs.benchmarks }}
         run: |
-          matrix=$(echo '[${{ github.event.inputs.benchmarks }}]' | jq '.[] | select(endswith("Benchmark")) | .')
+          matrix=$(echo "[$BENCHMARKS_INPUT]" | jq '.[] | select(endswith("Benchmark")) | .')
```
maluchari, line 53: Is this issue specific to this branch?

maluchari (approved):

.github/workflows/jmh-benchmarks.yml
vikrambohra: Should this be merged into master as well. I can do it as part of separate PR

shanthoosh (approved):

PR #233, opened 2026-03-04
Fix bin-pack compaction producing undersized output files

teamurko:
SparkStagedScan uses task.sizeBytes() (data + delete files) as the weight function for bin-packing, but the split size passed from SparkBinPackDataRewriter is computed using ContentScanTask::length (data-only). This mismatch causes each Spark partition to hold less actual data than intended, producing output files smaller than target-file-size-bytes.

Change SparkStagedScan to use ContentScanTask::length as the bin-packing weight, consistent with how splitSize was computed. Also fix a pre-existing bug in SparkStagedScan.hashCode() where splitSize was used twice instead of including splitLookback.

Fix task weight in budget truncation logic from sizeBytes to length.

### DIFF
--- a/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/SparkBinPackDataRewriter.java
+++ b/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/SparkBinPackDataRewriter.java
@@ -49,7 +49,6 @@ class SparkBinPackDataRewriter extends SparkSizeBasedDataRewriter {
             .option(SparkReadOptions.SCAN_TASK_SET_ID, groupId)
             .option(SparkReadOptions.SPLIT_SIZE, splitSize(inputSize(group)))
             .option(SparkReadOptions.FILE_OPEN_COST, "0")
-            .option(SparkReadOptions.USE_DATA_ONLY_WEIGHT, "true")
             .load(groupId);
 
     // write the packed data into new files where each split becomes a new file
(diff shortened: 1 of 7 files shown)


spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/actions/SparkBinPackDataRewriter.java
```diff
@@ -49,6 +49,7 @@ protected void doRewrite(String groupId, List<FileScanTask> group) {
             .option(SparkReadOptions.SCAN_TASK_SET_ID, groupId)
             .option(SparkReadOptions.SPLIT_SIZE, splitSize(inputSize(group)))
             .option(SparkReadOptions.FILE_OPEN_COST, "0")
+            .option(SparkReadOptions.USE_DATA_ONLY_WEIGHT, "true")
```
sumedhsakdeo: Is there a way to do without an additional knob like this? Make it default behavior

teamurko: discussed offline, https://github.com/linkedin/iceberg/blob/809534da01bc5a3002c5edd427e91eff09ec2443/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkTable.java#L270, sparkstagedscan is used only by rewrite procedures

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkStagedScan.java
```diff
@@ -59,7 +60,16 @@ protected List<ScanTaskGroup<ScanTask>> taskGroups() {
           table(),
           taskSetId);
 
-      this.taskGroups = TableScanUtil.planTaskGroups(tasks, splitSize, splitLookback, openFileCost);
+      Function<ScanTask, Long> weightFunc =
```
sumedhsakdeo: This feels like are leaking some internal details into engine layer. Can we keep the weightFunc in TableScanUtils with a reasonable default calculation?

teamurko: TableScanUtils.planFileGroups is used in other types of scan where the default should use sizeBytes. To hide the weight calculation we can add a new method to TableScanUtils

sumedhsakdeo: Yea that might be fine too.

teamurko: done

sumedhsakdeo (approved):

PR #237, opened 2026-04-15
Suppress file deletion errors during Spark write abort

teamurko:
Suppress file deletion errors during abort to prevent masking original commit failure

Previously, both the driver-level and task-level abort paths used throwFailureWhenFinished(), causing HDFS delete errors (e.g. RemoteException) to propagate to the user and mask the original commit failure. This also meant task-level deletes had no retries, so transient failures left orphan files.

Now both paths suppress delete exceptions (logging each at WARN level) and the task-level path retries up to 3 times with exponential backoff, aligning with the Spark 3.5 implementation.

This change is impl of the same behavior from iceberg1.5.2 for spark3.5: https://github.com/linkedin/iceberg/blob/c9c41c46fbf5102b657e53981b1ec534010338a7/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkCleanupUtil.java#L118

### DIFF
--- a/spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
+++ b/spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
@@ -20,14 +20,6 @@ package org.apache.iceberg.spark.source;
 
 import static org.apache.iceberg.IsolationLevel.SERIALIZABLE;
 import static org.apache.iceberg.IsolationLevel.SNAPSHOT;
-import static org.apache.iceberg.TableProperties.COMMIT_MAX_RETRY_WAIT_MS;
-import static org.apache.iceberg.TableProperties.COMMIT_MAX_RETRY_WAIT_MS_DEFAULT;
-import static org.apache.iceberg.TableProperties.COMMIT_MIN_RETRY_WAIT_MS;
-import static org.apache.iceberg.TableProperties.COMMIT_MIN_RETRY_WAIT_MS_DEFAULT;
-import static org.apache.iceberg.TableProperties.COMMIT_NUM_RETRIES;
-import static org.apache.iceberg.TableProperties.COMMIT_NUM_RETRIES_DEFAULT;
-import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS;
-import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -69,7 +61,6 @@ import org.apache.iceberg.relocated.com.google.common.collect.Sets;
 import org.apache.iceberg.spark.CommitMetadata;
 import org.apache.iceberg.spark.FileRewriteCoordinator;
 import org.apache.iceberg.spark.SparkWriteConf;
-import org.apache.iceberg.util.PropertyUtil;
 import org.apache.iceberg.util.Tasks;
 import org.apache.spark.TaskContext;
 import org.apache.spark.TaskContext$;
@@ -92,6 +83,11 @@ import org.slf4j.LoggerFactory;
 
 class SparkWrite {
   private static final Logger LOG = LoggerFactory.getLogger(SparkWrite.class);
+
+  private static final int DELETE_NUM_RETRIES = 3;
+  private static final int DELETE_MIN_RETRY_WAIT_MS = 100; // 100 ms
+  private static final int DELETE_MAX_RETRY_WAIT_MS = 30 * 1000; // 30 seconds
+  private static final int DELETE_TOTAL_RETRY_TIME_MS = 2 * 60 * 1000; // 2 minutes
 
   private final JavaSparkContext sparkContext;
   private final Table table;
@@ -244,18 +240,16 @@ class SparkWrite {
 
   private void abort(WriterCommitMessage[] messages) {
     if (cleanupOnAbort) {
-      Map<String, String> props = table.properties();
       Tasks.foreach(files(messages))
-          .retry(PropertyUtil.propertyAsInt(props, COMMIT_NUM_RETRIES, COMMIT_NUM_RETRIES_DEFAULT))
+          .retry(DELETE_NUM_RETRIES)
           .exponentialBackoff(
-              PropertyUtil.propertyAsInt(
-                  props, COMMIT_MIN_RETRY_WAIT_MS, COMMIT_MIN_RETRY_WAIT_MS_DEFAULT),
-              PropertyUtil.propertyAsInt(
-                  props, COMMIT_MAX_RETRY_WAIT_MS, COMMIT_MAX_RETRY_WAIT_MS_DEFAULT),
-              PropertyUtil.propertyAsInt(
-                  props, COMMIT_TOTAL_RETRY_TIME_MS, COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT),
+              DELETE_MIN_RETRY_WAIT_MS,
+              DELETE_MAX_RETRY_WAIT_MS,
+              DELETE_TOTAL_RETRY_TIME_MS,
               2.0 /* exponential */)
-          .throwFailureWhenFinished()
+          .suppressFailureWhenFinished()
+          .onFailure(
+              (file, exc) -> LOG.warn("Failed to delete {} during job abort", file.path(), exc))
           .run(
               file -> {
                 table.io().deleteFile(file.path().toString());
@@ -668,8 +662,15 @@ class SparkWrite {
 
   private static <T extends ContentFile<T>> void deleteFiles(FileIO io, List<T> files) {
     Tasks.foreach(files)
-        .throwFailureWhenFinished()
-        .noRetry()
+        .suppressFailureWhenFinished()
+        .retry(DELETE_NUM_RETRIES)
+        .exponentialBackoff(
+            DELETE_MIN_RETRY_WAIT_MS,
+            DELETE_MAX_RETRY_WAIT_MS,
+            DELETE_TOTAL_RETRY_TIME_MS,
+            2.0 /* exponential */)
+        .onFailure(
+            (file, exc) -> LOG.warn("Failed to delete {} during task abort", file.path(), exc))
         .run(file -> io.deleteFile(file.path().toString()));
   }
 
--- a/spark/v3.1/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkDataWrite.java
+++ b/spark/v3.1/spark/src/test/java/org/apache/iceberg/spark/source/TestSparkDataWrite.java
@@ -670,9 +670,7 @@ public class TestSparkDataWrite {
     // so cleanupOnAbort stays true and abort will attempt file deletion)
     AppendFiles append = table.newAppend();
     AppendFiles spyAppend = spy(append);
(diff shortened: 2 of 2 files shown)


spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -93,6 +84,11 @@
 class SparkWrite {
   private static final Logger LOG = LoggerFactory.getLogger(SparkWrite.class);
 
+  private static final int DELETE_NUM_RETRIES = 3;
```
dxichen, line 87: Thoughts about surfacing these as a tableproperties?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -668,8 +662,15 @@ public DataWriter<InternalRow> createWriter(int partitionId, long taskId, long e
...
-        .noRetry()
+        .suppressFailureWhenFinished()
+        .retry(DELETE_NUM_RETRIES)
+        .exponentialBackoff(
+            DELETE_MIN_RETRY_WAIT_MS,
+            DELETE_MAX_RETRY_WAIT_MS,
+            DELETE_TOTAL_RETRY_TIME_MS,
+            2.0 /* exponential */)
+        .onFailure(
+            (file, exc) -> LOG.warn("Failed to delete {} during task abort", file.path(), exc))
```
dxichen, line 673: This failure is not limited to abort but also table deletes for which case we should not swallow it

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -244,18 +240,16 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
...
-              PropertyUtil.propertyAsInt(
-                  props, COMMIT_TOTAL_RETRY_TIME_MS, COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT),
+              DELETE_MIN_RETRY_WAIT_MS,
+              DELETE_MAX_RETRY_WAIT_MS,
+              DELETE_TOTAL_RETRY_TIME_MS,
               2.0 /* exponential */)
-          .throwFailureWhenFinished()
+          .suppressFailureWhenFinished()
+          .onFailure(
+              (file, exc) -> LOG.warn("Failed to delete {} during job abort", file.path(), exc))
```
dxichen, line 252: Prefer to emit a metrics via a custom MetricsReporter here but could be handled in a separate future pr

teamurko: Yes as an option, but Iceberg relies on logs for these cases

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
teamurko: table drop makes the catalog server delete the table directory, it's a different code path

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
mkuchenbecker: +1, i'm fine these being default where not specified.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -244,18 +240,16 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
...
-              PropertyUtil.propertyAsInt(
-                  props, COMMIT_MAX_RETRY_WAIT_MS, COMMIT_MAX_RETRY_WAIT_MS_DEFAULT),
-              PropertyUtil.propertyAsInt(
-                  props, COMMIT_TOTAL_RETRY_TIME_MS, COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT),
+              DELETE_MIN_RETRY_WAIT_MS,
+              DELETE_MAX_RETRY_WAIT_MS,
+              DELETE_TOTAL_RETRY_TIME_MS,
               2.0 /* exponential */)
-          .throwFailureWhenFinished()
+          .suppressFailureWhenFinished()
```
mkuchenbecker, line 250: This line is the main change though, correct?

mkuchenbecker, line 250: Any way to disable / toggle in the event of an issue?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -244,18 +240,16 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
 
   private void abort(WriterCommitMessage[] messages) {
     if (cleanupOnAbort) {
-      Map<String, String> props = table.properties();
       Tasks.foreach(files(messages))
-          .retry(PropertyUtil.propertyAsInt(props, COMMIT_NUM_RETRIES, COMMIT_NUM_RETRIES_DEFAULT))
+          .retry(DELETE_NUM_RETRIES)
```
mkuchenbecker, line 244: What is the impact of not retrying? Stale data later GC'd?
What is the latency impact of retrying if storage is down?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -244,18 +240,16 @@ private void commitOperation(SnapshotUpdate<?> operation, String description) {
 
   private void abort(WriterCommitMessage[] messages) {
     if (cleanupOnAbort) {
```
mkuchenbecker, line 242: How often is cleanup on abort invoked? Do we have data?

mkuchenbecker (commented):
Generally seems fine. I want to know how we control it and if its worth retrying vs failing fast and just ignoring it.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
teamurko: i wouldn't deviate much from https://github.com/linkedin/iceberg/blob/c9c41c46fbf5102b657e53981b1ec534010338a7/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkCleanupUtil.java#L118
it's almost a back-port

dxichen (commented):
Thanks for the backport

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -20,14 +20,6 @@
...
 import static org.apache.iceberg.IsolationLevel.SERIALIZABLE;
 import static org.apache.iceberg.IsolationLevel.SNAPSHOT;
-import static org.apache.iceberg.TableProperties.COMMIT_MAX_RETRY_WAIT_MS;
-import static org.apache.iceberg.TableProperties.COMMIT_MAX_RETRY_WAIT_MS_DEFAULT;
-import static org.apache.iceberg.TableProperties.COMMIT_MIN_RETRY_WAIT_MS;
-import static org.apache.iceberg.TableProperties.COMMIT_MIN_RETRY_WAIT_MS_DEFAULT;
-import static org.apache.iceberg.TableProperties.COMMIT_NUM_RETRIES;
-import static org.apache.iceberg.TableProperties.COMMIT_NUM_RETRIES_DEFAULT;
-import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS;
-import static org.apache.iceberg.TableProperties.COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT;
```
abhisheknath2011, line 30: Are these properties removed?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -93,6 +84,11 @@
 class SparkWrite {
   private static final Logger LOG = LoggerFactory.getLogger(SparkWrite.class);
 
+  private static final int DELETE_NUM_RETRIES = 3;
+  private static final int DELETE_MIN_RETRY_WAIT_MS = 100; // 100 ms
+  private static final int DELETE_MAX_RETRY_WAIT_MS = 30 * 1000; // 30 seconds
+  private static final int DELETE_TOTAL_RETRY_TIME_MS = 2 * 60 * 1000; // 2 minutes
```
abhisheknath2011, line 90: Are these default or updated properties?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
teamurko: + retries using delete specific params instead of commit params

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
teamurko: these are used in other places

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
teamurko: exact values used for delete retry

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
abhisheknath2011: Not sure if we can emit metrics as we need to configure OTEL or emission lib?

mkuchenbecker (commented):
Main comments:

1. How do we turn it off given an issue? [rollout killswitch or knob]
2. Adding retries, is that needed? Given degradation of storage, would the request take longer now because its trying to delete?

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
jiang95-dev: Cleanup is quite often. Whenever the server sends `CommitFailedException` the engine does cleanup.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
jiang95-dev: I think we can add table properties like `write.abort.suppress-delete-failures` to toggle this in production.

spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
jiang95-dev: I think in the issue like running out of HDFS inode, the executors are more likely to fail, and so there's no point of retrying. But normally if it is just a commit failure, the driver should clean up with retries. So we can keep the current setting to only let the driver retry.

PR #238, opened 2026-04-30
Suppress file deletion errors during Spark write abort

jiang95-dev:
This PR copies https://github.com/linkedin/iceberg/pull/237, and addressed the comments by changing 2 things:

1. Add toggle on the suppress. User can configure the table property to suppress or unsuppress the error message.
2. Remove retries from the executor abort to avoid burden / latency increase to hdfs during issues. Leave the driver abort with retries.

### DIFF
--- a/spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
+++ b/spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
@@ -20,14 +20,6 @@ package org.apache.iceberg.spark.source;
 
 import static org.apache.iceberg.IsolationLevel.SERIALIZABLE;
 import static org.apache.iceberg.IsolationLevel.SNAPSHOT;
-import static org.apache.iceberg.TableProperties.COMMIT_MAX_RETRY_WAIT_MS;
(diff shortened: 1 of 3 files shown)


spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java
```diff
@@ -93,6 +87,11 @@
 class SparkWrite {
   private static final Logger LOG = LoggerFactory.getLogger(SparkWrite.class);
 
+  private static final int DELETE_NUM_RETRIES = 3;
```
mkuchenbecker, line 90: What is the value in retrying? Does the deletion happen synchronously or async? My main thought is the deletion will succeed 98% of the time, and the other 2% are likely during times when the cluster is degraded and retrying will not work. 

Not retrying we delete 98% of the data, the other 2% are GC'd async by OFD. If we retry we need to confirm there is no impact to the customer performance, but I don't see a strong value given OFD is a backstop.

mkuchenbecker (approved):
Approving as this is a backport, essentially. My suggestion is we add a config option for number of retries, as it happens synchronously and the impact of failure to delete is backstopped by OFD.

abhisheknath2011 (approved):

PR #243, opened 2026-05-06
Backport ExtendedV2Writes rule for Spark 3.1

chenselena:
## Summary                                                                                                                                                                                                    
  Backports Spark 3.5's V2Writes idea to Spark 3.1: a rule that attaches a local Sort to V2 write commands (`AppendData`/`OverwriteByExpression`/`OverwritePartitionsDynamic`) so partitioned writes don't fail
  in `ClusteredDataWriter`. Aligns Spark 3.1 behavior with Spark 3.5 ahead of the upgrade.                                                                                                                      
   
  ## Behavior                                                                                                                                                                                                   
  - **Sort attached** when the table has an explicit sort order (partition cols prepended via `Spark3Util.buildRequiredOrdering`).
  - **No sort attached** for unsorted tables — users must enable `write.spark.fanout.enabled=true` or pre-cluster the input. Matches Spark 3.5's default path.                                                  
  - **No Exchange** is ever attached: Spark 3.1 only has strict `RepartitionByExpression`, and a strict repartition would turn skewed partition keys into stragglers. Spark 3.4+'s non-strict                   
  `RebalancePartitions` (which Spark 3.5 uses) doesn't exist here.                                                                                                                                              
  - **MERGE/UPDATE/DELETE skipped** — `RewriteRowLevelOperationHelper.buildWritePlan` already prepares those queries; `alreadyPrepared()` detects its output shape to avoid double-wrapping. 
                                                            
  ## Files                                                                                                                                                                                                      
  - `ExtendedV2Writes.scala` — the rule + a local `requiredOrdering` wrapper around `Spark3Util.buildRequiredOrdering`.
  - `IcebergSparkSessionExtensions.scala` — registers the rule.                                                                                                                                                 
  - `TestRequiredDistributionAndOrdering.java` — covers sorted/RANGE/HASH/NONE distributions, fanout paths, AQE, single-shuffle-partition, null partition values, empty input, and `saveAsTable` append.        
  `*FailsWithoutRule` tests pin the pre-rule baseline.                                                                                                                                                          
                                                                                                                                                                                                                
  ## Test plan                                                                                                                                                                                                  
  - [x] `./gradlew :iceberg-spark:iceberg-spark-extensions-3.1_2.12:test --tests "*TestRequiredDistributionAndOrdering*"`

### DIFF
--- a/spark/v3.1/spark-extensions/src/main/scala/org/apache/iceberg/spark/extensions/IcebergSparkSessionExtensions.scala
+++ b/spark/v3.1/spark-extensions/src/main/scala/org/apache/iceberg/spark/extensions/IcebergSparkSessionExtensions.scala
@@ -31,6 +31,7 @@ import org.apache.spark.sql.catalyst.optimizer.RewriteMergeInto
 import org.apache.spark.sql.catalyst.optimizer.RewriteUpdate
 import org.apache.spark.sql.catalyst.parser.extensions.IcebergSparkSqlExtensionsParser
 import org.apache.spark.sql.execution.datasources.v2.ExtendedDataSourceV2Strategy
+import org.apache.spark.sql.execution.datasources.v2.ExtendedV2Writes
 
 class IcebergSparkSessionExtensions extends (SparkSessionExtensions => Unit) {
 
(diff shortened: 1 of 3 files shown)


dxichen (approved):
thanks for the backport!

dxichen (approved):

maluchari (approved):

PR #249, opened 2026-06-16
Spark 3.5: Default partitioned write distribution mode to NONE

shardulm94:
## Summary

Iceberg 1.5 implements `RequiresDistributionAndOrdering`, so every Spark 3.5 write to a partitioned table now requests a **HASH** distribution by default (`SparkWriteConf.defaultWriteDistributionMode`). Spark plans this as a `RebalancePartitions` / REBALANCE shuffle on the partition columns.

REBALANCE is poorly handled by **Celeborn 0.4** and has already caused production incidents (notably Kyoto CDC writes overloading Celeborn and impacting other tenants). Celeborn 0.6 fixes this but is still a couple of months out, while Spark 3.5 traffic is about to ramp significantly with the Python 3.12 migration this half.

This PR flips the default write distribution mode for **partitioned, unsorted** tables from `HASH` to `NONE`, restoring the Spark 3.1 status quo:

- **Scope is intentionally narrow.** Only normal (batch/append) writes are affected. `DELETE`/`UPDATE`/`MERGE` keep their existing `HASH` defaults, which covers the CDC merge-into case separately.
- **Ordering is unchanged.** We keep the local sort, so users don't need to add `.sortWithinPartitions()` in their code.
- **Explicit requests still work.** Users who want a distribution can still request it via the write option, session conf, or table property.
- Any small files that result are handled by Open House's post-write compaction. We can revisit re-enabling the default distribution once Celeborn 0.6 is rolled out.

This also avoids the Iceberg-injected shuffle silently overriding user intent (e.g. `DaliSpark.OUTPUT_PARALLELISM`) and the extra resource usage it adds.

## Testing Done

- [x] Local code review completed
- [x] Unit tests updated to reflect the new default
- [x] Targeted test suites pass on JDK 11

Updated tests:
- `TestSparkWriteConf#testSparkWriteConfDistributionDefault` — now asserts `NONE` for normal writes and `HASH` for row-level operations (DELETE/UPDATE/MERGE).
- `TestSparkWriteConf#testAdvisoryPartitionSize` — explicitly requests `HASH` so an advisory partition size is reported (Spark only reports one when a distribution is requested).
- `TestSparkDistributionAndOrderingUtil#testDefaultWritePartitionedUnsortedTable` — now expects an unspecified distribution with the local ordering retained.

Ran `TestSparkWriteConf`, `TestSparkDistributionAndOrderingUtil`, and `TestRequiredDistributionAndOrdering` (both `spark` and `spark-extensions` modules) — all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java
+++ b/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java
@@ -316,7 +316,7 @@ public class SparkWriteConf {
     if (table.sortOrder().isSorted()) {
       return RANGE;
     } else if (table.spec().isPartitioned()) {
-      return HASH;
+      return NONE;
     } else {
       return NONE;
     }
--- a/spark/v3.5/spark/src/test/java/org/apache/iceberg/spark/TestSparkDistributionAndOrderingUtil.java
+++ b/spark/v3.5/spark/src/test/java/org/apache/iceberg/spark/TestSparkDistributionAndOrderingUtil.java
@@ -125,8 +125,11 @@ public class TestSparkDistributionAndOrderingUtil extends TestBaseWithCatalog {
   //
   // PARTITIONED BY date, days(ts) UNORDERED
   // -------------------------------------------------------------------------
-  // write mode is NOT SET -> CLUSTER BY date, days(ts) + LOCALLY ORDER BY date, days(ts)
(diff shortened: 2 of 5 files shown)


cbb330:
Thanks for the change. Discussed offline on the implications and it will not have impact to largest customers and keeps parity with spark 3.1.

PR #263, opened 2026-07-23
ORC: fill initial defaults for missing id-bound fields

cbb330:
## Summary

Fills a field's `initial-default` when it is absent from an id-bearing ORC file, using the same fallback order as the Parquet reader.

- `buildOrcProjection` omits an absent defaulted field only when the file has embedded IDs **and the configured reader explicitly declares `supportsInitialDefaults()`**.
- The id-binding `StructReader` from #265 then materializes the declared default as a per-file constant, consuming no column vector.
- A present column always wins, including explicit null. ID-less/name-mapped files keep legacy null synthesis.
- Readers that do not opt in retain their previous projection shape and behavior.
- Spark 3.1 scans projecting a default at any nesting level are routed away from vectorized ORC reads; default materialization remains iterative-only.

## Why reader opt-in is required

`OrcIterable` is shared by Generic, Spark, and Flink row and vectorized readers. Embedded field IDs establish that an absent column was never written, but do not establish that the selected reader can fill it. The new `ORC.ReadBuilder.supportsInitialDefaults()` capability keeps omission local to readers that implement the missing-field contract.

Generic row reads opt in here. Spark 3.1 row reads opt in in #264. This branch also provides the recursive Spark 3.1 scan-routing guard required before the row-reader implementation is enabled. Other engines and vectorized readers do not opt in.

## Provenance

This is net new rather than a backport. Apache Iceberg 1.2 has the schema model for defaults but does not implement ORC default reads. The resolution logic mirrors Parquet's field-reader fallback: physical reader → declared default → null/failure.

## Supported scope

Generic ORC row reader; scalar defaults at any struct nesting level, including structs inside list elements and map values.

Vectorized ORC default materialization remains deferred. Predicates on omitted defaulted fields use a conservative ORC SearchArgument so they are evaluated after iterative default materialization.

## Stack

1. #265 — ID-based `StructReader` backport
2. **This PR** — Generic ORC initial-default fill, reader capability gate, and Spark 3.1 iterative-routing guard
3. #264 — Spark 3.1 ORC initial-default fill

## Testing Done
- [x] Projection tests cover opted-in embedded-ID omission and non-opted null synthesis
- [x] Generic ORC end-to-end coverage for scalar, required, nested, list, and map defaults
- [x] Present values and explicit nulls override defaults
- [x] ID-less files retain null behavior
- [x] Top-level and nested projected defaults disable Spark 3.1 vectorized ORC reads
- [x] Schemas without projected defaults remain vectorization eligible
- [x] Full `iceberg-orc` and `iceberg-data` test suites pass
- [x] Spotless passes

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/ORC.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORC.java
@@ -672,6 +672,7 @@ public class ORC {
 
     private Function<TypeDescription, OrcRowReader<?>> readerFunc;
     private Function<TypeDescription, OrcBatchReader<?>> batchedReaderFunc;
+    private boolean supportsInitialDefaults = false;
     private int recordsPerBatch = VectorizedRowBatch.DEFAULT_SIZE;
 
     private ReadBuilder(InputFile file) {
@@ -725,6 +726,19 @@ public class ORC {
       return this;
     }
 
+    /**
+     * Signals that the configured row reader can fill initial defaults for fields omitted from an
+     * ORC file. Disabled by default so existing positional and vectorized readers retain their
+     * previous null-synthesizing projection behavior.
+     */
+    public ReadBuilder supportsInitialDefaults() {
+      Preconditions.checkState(
+          this.readerFunc != null,
+          "A row reader function must be configured before enabling initial defaults");
+      this.supportsInitialDefaults = true;
+      return this;
+    }
+
     public ReadBuilder filter(Expression newFilter) {
       this.filter = newFilter;
       return this;
@@ -733,7 +747,7 @@ public class ORC {
     public ReadBuilder createBatchedReaderFunc(
         Function<TypeDescription, OrcBatchReader<?>> batchReaderFunction) {
       Preconditions.checkArgument(
-          this.readerFunc == null,
+          this.readerFunc == null && !this.supportsInitialDefaults,
           "Batched reader function cannot be set since the non-batched version is already set");
       this.batchedReaderFunc = batchReaderFunction;
       return this;
@@ -759,6 +773,7 @@ public class ORC {
           start,
           length,
           readerFunc,
+          supportsInitialDefaults,
           caseSensitive,
           filter,
           batchedReaderFunc,
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -47,6 +47,32 @@ public final class ORCSchemaUtil {
     LONG
   }
 
+  /**
+   * Where the Iceberg field IDs in an ORC schema came from.
+   *
+   * <p>This is the provenance of the IDs, not a statement about the file's contents. It decides
+   * whether the absence of an ID may be read as "this column was never written", which in turn
+   * decides whether a declared {@code initial-default} may be filled on read.
+   */
+  enum FieldIdSource {
+    /**
+     * The IDs were read from {@code iceberg.id} column attributes written into the file by {@link
+     * ORCSchemaUtil#convert(Schema)}. A field with no ID was genuinely never written, so a declared
+     * default may be filled.
+     */
+    EMBEDDED,
+
+    /**
+     * The IDs were derived at read time by {@link ORCSchemaUtil#applyNameMapping} matching column
+     * names, because the file carried none of its own -- typically a legacy or Hive-migrated file.
+     * A field can look absent merely because its name did not match, for example after a rename the
+     * name mapping no longer covers, while its data is physically present in the file. Filling a
+     * default here would fabricate values over real data, so absent fields are synthesized as null
+     * columns instead.
+     */
+    NAME_MAPPED
+  }
+
   private static class OrcField {
     private final String name;
     private final TypeDescription type;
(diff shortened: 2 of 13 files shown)


orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -261,18 +261,47 @@ public static Schema convert(TypeDescription orcSchema) {
...
+   * projection. An id-binding reader sees no column for that field and fills the declared default
+   * as a per-file constant. When false, the field is synthesized as a null column instead, so a
+   * default is never name-matched onto an id-less file.
+   */
+  static TypeDescription buildOrcProjection(
+      Schema schema, TypeDescription originalOrcSchema, boolean hasTrustedIds) {
     final Map<Integer, OrcField> icebergToOrc = icebergToOrcMapping("root", originalOrcSchema);
-    return buildOrcProjection(Integer.MIN_VALUE, schema.asStruct(), true, icebergToOrc);
+    return buildOrcProjection(
+        Integer.MIN_VALUE, schema.asStruct(), true, hasTrustedIds, icebergToOrc);
```
mkuchenbecker: what are trusted IDs?

orc/src/main/java/org/apache/iceberg/orc/OrcValueReaders.java
```diff
@@ -208,6 +233,10 @@ protected StructReader(
           this.isConstantOrMetadataField[pos] = false;
           this.orcFieldIndex[pos] = fieldIdToOrcIndex.getOrDefault(field.fieldId(), -1);
           this.readers[pos] = fileReader;
+        } else if (convertConstant != null && field.initialDefault() != null) {
+          this.isConstantOrMetadataField[pos] = true;
+          this.readers[pos] =
```
mkuchenbecker, line 238: actual logic

mkuchenbecker (commented):
overall makes sense, datatypes tested, where there is a default and the default is not specified, then populate. 

An intergration test in li-openhouse for trino and spark will be useful to capture actual behaviour.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
cbb330: name is a bit vague so I changed it to a two-valued enum, which names where the field ids in the schema came from.

- EMBEDDED: the ids were read from iceberg.id column attributes persisted in the ORC file. When it is present it is great, a field with no id was genuinely never written, so its declared default can be filled safely.
- NAME_MAPPED: the file carried no ids of its own (e.g. was renamed from kafka ETL's Hive-migration), so an earlier step derived them at read time by matching column names. When file carries no ids, "absent" is not proof since e.g. an iceberg column can rename and the name mapping no longer covers.but it's data is physically in the file. Filling a default there would fabricate values over real data, so we don't fill the default on this path. We keep the original behavior here, in order to focus only on the important case: iceberg created files.

cbb330:
I added an explicit `supportsInitialDefaults()` flag to the ORC row-reader path.

Previously, the shared ORC projection omitted missing defaulted columns whenever a file had embedded field IDs. That inadvertently affected every Spark/Flink row and vectorized reader, even though only Generic and Spark 3.1 iterative readers know how to fill the omitted column today. This could break non-supported engines by giving their positional readers a shorter column list.

 Now a field is omitted only when all four conditions hold:

 ```java
   supportsInitialDefaults
       && field.initialDefault() != null
       && !mapping.containsKey(field.fieldId())
       && fieldIdSource == FieldIdSource.EMBEDDED
 ```

 Generic and Spark 3.1 iterative readers explicitly opt in. Other engines retain the previous projection and behavior. Spark 3.1 also avoids vectorized ORC reads when the projected schema contains
 initial defaults; vectorized support can be added separately.

This keeps default filling isolated to readers that implement it, while preserving backward compatibility everywhere else.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -47,6 +47,32 @@ public enum LongType {
...
+
+    /**
+     * The IDs were derived at read time by {@link ORCSchemaUtil#applyNameMapping} matching column
+     * names, because the file carried none of its own -- typically a legacy or Hive-migrated file.
+     * A field can look absent merely because its name did not match, for example after a rename the
+     * name mapping no longer covers, while its data is physically present in the file. Filling a
+     * default here would fabricate values over real data, so absent fields are synthesized as null
+     * columns instead.
+     */
+    NAME_MAPPED
```
mkuchenbecker, line 73: Why are we adding name mapping? The renamed files from Kyoto? This worries me building in the implicit mapping. If we didn't add the legacy mapping, what happens? Those files are unreadable?

mkuchenbecker: I think a restriction to use this to keep it simple we could require a re-write of the table so it has headers.

orc/src/main/java/org/apache/iceberg/orc/ORC.java
```diff
@@ -725,6 +726,19 @@ public ReadBuilder createReaderFunc(Function<TypeDescription, OrcRowReader<?>> r
       return this;
     }
 
+    /**
+     * Signals that the configured row reader can fill initial defaults for fields omitted from an
+     * ORC file. Disabled by default so existing positional and vectorized readers retain their
+     * previous null-synthesizing projection behavior.
```
mkuchenbecker, line 732: Can we throw an error vs defaulting to null? this would avoid the correctness issue.

cbb330:
Superseded by #267 (omit + `idToConstant` on the Spark row reader). This was the Option A Generic fill path; we are not taking that design. Closing.

PR #267, opened 2026-08-10
[default values] ORC/Spark: forward-port default-value reads via idToConstant (LI #76)

cbb330:
## Summary

Forward-port of the Spark ORC default-value read path from [LI #76](https://github.com/linkedin/iceberg/pull/76) onto `openhouse-1.2.0`, adapted to [`initial-default`](https://github.com/linkedin/iceberg/pull/250) and stacked on [id-bound ORC structs (#265)](https://github.com/linkedin/iceberg/pull/265).

Absent defaulted fields are omitted from the per-file ORC projection and injected into `idToConstant`. The Spark **row** reader fills them. Omit is opt-in (`supportsInitialDefaults`) because `ORCSchemaUtil` is shared with Generic.

**Vectorized ORC default-fill is out of scope.** `SparkBatchScan` disables columnar ORC when the projection includes an `initial-default`. The vectorized-dir leftovers from #76 were dropped from this PR.

**Mechanism (same as #76):**
1. Omit absent defaulted fields from the per-file ORC projection
2. Inject their defaults into `idToConstant`
3. Existing `ConstantReader` materializes them

Inline notes mark each site as **as-is** or **changed** vs #76 ([review](https://github.com/linkedin/iceberg/pull/267#pullrequestreview-4920871589)).

Stack [#270](https://github.com/linkedin/iceberg/pull/267): this PR → [#268](https://github.com/linkedin/iceberg/pull/268) → [#269](https://github.com/linkedin/iceberg/pull/269).

---

### As-is

| Behavior | Where | Note |
|---|---|---|
| Overridable `visitRecord` | [`OrcSchemaWithTypeVisitor`](https://github.com/linkedin/iceberg/blob/chbush/oh120-orc-defaults-raymond-pr1/orc/src/main/java/org/apache/iceberg/orc/OrcSchemaWithTypeVisitor.java) | [as-is](https://github.com/linkedin/iceberg/pull/267#discussion_r3770102354) |
| Iceberg/ORC id alignment | [`containsInOrder`](https://github.com/linkedin/iceberg/blob/chbush/oh120-orc-defaults-raymond-pr1/spark/v3.1/spark/src/main/java/org/apache/iceberg/spark/OrcSchemaWithTypeVisitorSpark.java) | [as-is](https://github.com/linkedin/iceberg/pull/267#discussion_r3770102361) |
[...]

PR #268, opened 2026-08-12
ORC: omit initial-defaults only for complete embedded field IDs

cbb330:
## Summary

- Tighten PR1's omit path: a defaulted field is omitted from the ORC projection only when the reader opted in (`supportsInitialDefaults`), the file uses **EMBEDDED** Iceberg field IDs, and the file has a **complete** id set.
- Name-mapped or partially annotated files cannot prove a column was never written, so they keep the null-synthesizing path instead of filling a default.
- Stacked on [#267](https://github.com/linkedin/iceberg/pull/267) (Raymond forward-port).

## Testing Done

- [x] Unit tests added/updated (`TestBuildOrcProjection`, `TestORCSchemaUtil.testHasAllIds`, Spark reader partial-ID cases)
- [ ] Integration tests pass
- [ ] Manual testing performed

```bash
export JAVA_HOME=$(/usr/libexec/java_home -v 11)
./gradlew -DsparkVersions=3.1 -DscalaVersion=2.12 -DhiveVersions= -DflinkVersions= \
  :iceberg-orc:test --tests 'org.apache.iceberg.orc.TestBuildOrcProjection' \
  --tests 'org.apache.iceberg.orc.TestORCSchemaUtil' \
  :iceberg-spark:iceberg-spark-3.1_2.12:test \
  --tests 'org.apache.iceberg.spark.data.TestSparkOrcReaderForFieldsWithDefaultValue' \
  -x generateGitProperties
```

## Dependencies

- Requires [#267](https://github.com/linkedin/iceberg/pull/267).

### DIFF
--- a/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
+++ b/orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
@@ -47,6 +47,32 @@ public final class ORCSchemaUtil {
     LONG
   }
 
+  /**
+   * Where the Iceberg field IDs in an ORC schema came from.
+   *
+   * <p>This is the provenance of the IDs, not a statement about the file's contents. It decides
+   * whether the absence of an ID may be read as "this column was never written", which in turn
+   * decides whether a declared {@code initial-default} may be filled on read.
+   */
+  enum FieldIdSource {
+    /**
+     * The IDs were read from {@code iceberg.id} column attributes written into the file by {@link
+     * ORCSchemaUtil#convert(Schema)}. A field with no ID was genuinely never written, so a declared
+     * default may be filled.
+     */
+    EMBEDDED,
+
+    /**
+     * The IDs were derived at read time by {@link ORCSchemaUtil#applyNameMapping} matching column
+     * names, because the file carried none of its own — typically a legacy or Hive-migrated file. A
+     * field can look absent merely because its name did not match, for example after a rename the
+     * name mapping no longer covers, while its data is physically present in the file. Filling a
+     * default here would fabricate values over real data, so absent fields are synthesized as null
+     * columns instead.
+     */
+    NAME_MAPPED
+  }
+
   private static class OrcField {
     private final String name;
     private final TypeDescription type;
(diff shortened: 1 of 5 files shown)


cbb330:
@mkuchenbecker carrying over your notes from #263. they apply here more than on the closed PR (that was a discarded strategy. it was net-new code and too complex. here we are building off a previous implementation that landed in prod for hive, and forward porting the mechanism)

TLDR; we do not fill defaults on name-mapped files. We do that using`FieldIdSource` which represents where the IDs already on the read schema came from:

- `EMBEDDED`: the file wrote `iceberg.id` attributes. A missing ID means that column was never written, so it is safe to omit and fill.
- `NAME_MAPPED`: IDs were applied at read time by the existing Iceberg name mapping (e.g for Hive-migrated / no headers). These are not safe. e.g. After a rename the mapping does not cover it, a column can look absent while its data is still in the file. Filling a default there would overlay real data, so we never omit. We fall back to same null-synthesizing projection as today.

When we skip name mapping entirely, those files would still read the way they do today. We just would not fill defaults. That matches the restriction you suggested: only fill when the file already has complete embedded field IDs. Reading does not require a rewrite. Filling does, the file has to have been written (or rewritten) with Iceberg ORC headers. We can call that out in the as a feature contract. 

btw we did not throw. Throwing on name-mapped / incomplete-ID files would break existing reads. The correctness fix is that we do not fill unless `EMBEDDED` + `hasAllIds`. Otherwise keep today's null column. The Spark row reader that opted in, on a file with complete IDs, gets the default rather than null.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
```diff
@@ -47,6 +47,32 @@ public enum LongType {
...
+
+    /**
+     * The IDs were derived at read time by {@link ORCSchemaUtil#applyNameMapping} matching column
+     * names, because the file carried none of its own — typically a legacy or Hive-migrated file. A
+     * field can look absent merely because its name did not match, for example after a rename the
+     * name mapping no longer covers, while its data is physically present in the file. Filling a
+     * default here would fabricate values over real data, so absent fields are synthesized as null
+     * columns instead.
+     */
+    NAME_MAPPED
```
cbb330, line 73: @mkuchenbecker this is the site for your name-mapping / Kyoto / rewrite notes from #263.

We are not adding a mapping. `NAME_MAPPED` means IDs were applied at read time by the *existing* Iceberg name mapping, and we **never omit/fill** on that path (a rename the mapping misses can look absent while data is still in the file).

Fill only for `EMBEDDED` + complete IDs. Files without headers stay readable as today (null, not default). Full write-up: https://github.com/linkedin/iceberg/pull/268#issuecomment-5273541265

cbb330 (commented):
Inline pointer for the name-mapping / rewrite notes from #263 — full answer is in the PR conversation.

orc/src/main/java/org/apache/iceberg/orc/ORCSchemaUtil.java
mkuchenbecker: Do we need to handle this case or just enforce this as an invariant that all files must be embedded to enable?

cbb330: agreed, embedded fieldids are required to enable. its the enable invariant. We don't ever fill name-mapped files.

This branch is the guard, not extra handling. name-mapped stays on today's null path, and we use this flag to check for it up front.

mkuchenbecker (approved):
WE agreed that name map is not requried but was ported back becaus the code already handles it upstream. Existing cases we are targeting have IDs.

abhisheknath2011 (approved):
Approving to unblock the merge.

abhisheknath2011 (approved):

PR #269, opened 2026-08-12
[default values] ORC: don't push file filters for omitted initial-default fields

cbb330:
## Summary

When a defaulted column is omitted from the ORC file projection ([#267](https://github.com/linkedin/iceberg/pull/267)), Spark may still filter on it (`WHERE country = 'US'`). Pushing that filter into the file fails in two ways:

1. If the projection is *only* that omitted column, the read schema is empty and converting the filter crashes.
2. If the filter is pushed anyway, the file never had the column, so the predicate cannot see the filled default. Spark would filter the wrong thing.

This PR stops pushing file-level predicates for omitted defaulted fields. Spark evaluates those filters on the filled values after read. Empty projections no longer crash.

Also: `TypeUtil` schema copies (`AssignFreshIds` / `PruneColumns` / `ReassignIds`) now keep `initialDefault` via `NestedField.from(field)`, so table-create and `SELECT` round-trips do not drop defaults.

Stacked on [#268](https://github.com/linkedin/iceberg/pull/268) → [#267](https://github.com/linkedin/iceberg/pull/267).

## Testing Done

```bash
export JAVA_HOME=$(/usr/libexec/java_home -v 11)
./gradlew -DsparkVersions=3.1 -DscalaVersion=2.12 -DhiveVersions= -DflinkVersions= \
  :iceberg-api:test --tests 'org.apache.iceberg.types.TestTypeUtil' \
  :iceberg-orc:test --tests 'org.apache.iceberg.orc.TestExpressionToSearchArgument' \
  :iceberg-spark:iceberg-spark-3.1_2.12:test \
  --tests 'org.apache.iceberg.spark.source.TestSparkOrcInitialDefaultScan' \
  --tests 'org.apache.iceberg.spark.data.TestSparkOrcReaderForFieldsWithDefaultValue' \
  -x generateGitProperties
```

- Filter on only the omitted defaulted column — no crash, Spark sees the filled value
- Mixed physical + defaulted predicates
- Defaulted scans stay on the row reader even with vectorization enabled
- `initialDefault` survives TypeUtil copies

### DIFF
--- a/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java
+++ b/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java
@@ -88,11 +88,7 @@ class AssignFreshIds extends TypeUtil.CustomOrderSchemaVisitor<Type> {
     for (int i = 0; i < length; i += 1) {
       Types.NestedField field = fields.get(i);
       Type type = types.next();
-      if (field.isOptional()) {
-        newFields.add(Types.NestedField.optional(newIds.get(i), field.name(), type, field.doc()));
-      } else {
-        newFields.add(Types.NestedField.required(newIds.get(i), field.name(), type, field.doc()));
-      }
+      newFields.add(Types.NestedField.from(field).withId(newIds.get(i)).ofType(type).build());
     }
 
     return Types.StructType.of(newFields);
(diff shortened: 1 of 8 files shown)


mkuchenbecker (commented):
Part of the coordinated default-values review. The predicate change is correctness-preserving: when an initial-default column is omitted from the per-file ORC projection, the search argument returns `YES_NO_NULL` for that field, so ORC does not discard rows before Spark evaluates the filled value. Physical columns retain their pushdown leaves, and mixed predicates still push the safe physical part.

The nested empty-projection cases and Spark-visible filtering are covered. The schema-copy edits are consistent with upstream builder behavior and do not introduce a demonstrated regression in the cumulative tree.

I found no inline issues.

mkuchenbecker (approved):
LGTM. Looked good to me initially and nothing flagged in automated review.

abhisheknath2011 (approved):
LGTM

PR #272, opened 2026-08-17
Fix uncompressed ORC seek at logical stream end

byronellis:
## Summary
- Fix uncompressed ORC seeks to a logical stream end that coincides with a non-final disk-range boundary.
- Package the pinned ORC overlay into the Spark 3.1 shaded runtime while preserving upstream behavior outside the targeted seek condition.

## Testing Done
- [x] Local code review completed
- [x] Spark 3.1 ORC integration tests pass (2 tests, 0 failures, 0 errors)
- [x] Packaged overlay bytecode compared against upstream ORC 1.8.2-nohive
- [x] Shaded caller/callee protobuf descriptors and packaged class inventory verified

🤖 Generated with [GitHub Copilot CLI](https://docs.github.com/copilot/github-copilot-cli)

### DIFF
--- a/spark/v3.1/spark-runtime/src/integration/java/org/apache/iceberg/spark/TestOrcUncompressedSeek.java
+++ b/spark/v3.1/spark-runtime/src/integration/java/org/apache/iceberg/spark/TestOrcUncompressedSeek.java
@@ -7,37 +7,36 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *   http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
  */
 package org.apache.iceberg.spark;
-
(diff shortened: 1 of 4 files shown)


spark/v3.1/spark-runtime/src/main/java/org/apache/orc/impl/InStream.java
```diff
@@ -0,0 +1,902 @@
...
+          // Accept this stream's logical end even when later disk ranges belong to other streams.
+          boolean isLogicalEnd = desired == length && positionFile == curRange.getEnd();
+          if (curRange.getOffset() <= positionFile
+              && (isLogicalEnd
+                  || (curRange.next == null
+                      ? positionFile <= curRange.getEnd()
+                      : positionFile < curRange.getEnd()))) {
+            position = desired;
+            setCurrent(curRange, true);
+            return;
```
mkuchenbecker, line 221: Actual change.

mkuchenbecker (approved):

spark/v3.1/spark-runtime/src/integration/java/org/apache/iceberg/spark/TestOrcUncompressedSeek.java
```diff
@@ -0,0 +1,79 @@
+/*
```
rdsr, line 1: Can we articulate how with the fat jar be build, if this compiled src file comes after orc files it will overwrite [that's correct I presume] , if it comes before,  will the old orc files overwrite?

byronellis: If you can show me where the fat jar is built I can take a look and/or test the actual integration

byronellis: FWIW we can make this change pretty much anywhere we like... including after relocation if we like.

rdsr: the fat jar is triggered through the shadowJar gradle task I think (it must be in the build.gradle files). As long as this src file is the final file in the fat jar . I'm ok with it. Just wanted to highlight a potential issue

PR #277, opened 2026-09-09
Backport rewrite delete file fix (#10020) to iceberg 1.5

jiang95-dev:
## Summary
Backport the fix from [iceberg #10020](https://github.com/apache/iceberg/pull/10020). This is needed because we ran into the same issue in production. 

## Testing Done
gradle build.

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -1018,6 +1018,16 @@ acceptedBreaks:
       old: "method void org.apache.iceberg.PositionDeletesTable.PositionDeletesBatchScan::<init>(org.apache.iceberg.Table,\
         \ org.apache.iceberg.Schema, org.apache.iceberg.TableScanContext)"
       justification: "Removing deprecated code"
+  "1.5.2.21":
+    com.linkedin.iceberg:iceberg-api:
+    - code: "java.class.defaultSerializationChanged"
+      old: "class org.apache.iceberg.PartitionSpec"
+      new: "class org.apache.iceberg.PartitionSpec"
+      justification: "Serialization across versions is not supported"
+    - code: "java.class.defaultSerializationChanged"
+      old: "class org.apache.iceberg.Schema"
+      new: "class org.apache.iceberg.Schema"
+      justification: "Serialization across versions is not supported"
   "1.5.2.14":
     com.linkedin.iceberg:iceberg-api:
     - code: "java.class.defaultSerializationChanged"
--- a/api/src/main/java/org/apache/iceberg/types/AssignIds.java
+++ b/api/src/main/java/org/apache/iceberg/types/AssignIds.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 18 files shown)


.palantir/revapi.yml
```diff
@@ -1018,6 +1018,16 @@ acceptedBreaks:
...
+  "1.5.2.21":
+    com.linkedin.iceberg:iceberg-api:
+    - code: "java.class.defaultSerializationChanged"
+      old: "class org.apache.iceberg.PartitionSpec"
+      new: "class org.apache.iceberg.PartitionSpec"
+      justification: "Serialization across versions is not supported"
+    - code: "java.class.defaultSerializationChanged"
+      old: "class org.apache.iceberg.Schema"
+      new: "class org.apache.iceberg.Schema"
+      justification: "Serialization across versions is not supported"
```
mkuchenbecker, line 1030: I am unfamiliar with .palantir revapi

What is this checking?

mkuchenbecker:
what version are we backporting from?

mkuchenbecker:
Its unclear from your commits what is backported vs fix forward. I might backport and commit merge conflicts, then update forward with new commits.

jiang95-dev:
> what version are we backporting from?

1.6.0

jiang95-dev:
> Its unclear from your commits what is backported vs fix forward. I might backport and commit merge conflicts, then update forward with new commits.

Currently there are 2 commits. First for backport. Second for fix forward.

.palantir/revapi.yml
jiang95-dev: RevAPI is the Java API/ABI compatibility check run by the `API Binary Compatibility Checks` CI job (`./gradlew revapi`). It compares the current `iceberg-api` JAR with the last published 1.5.2 baseline.

This backport adds public ID-remapping APIs to `Schema` and `PartitionSpec`. Because both classes are `Serializable` and do not declare an explicit `serialVersionUID`, those additions change the JVM-computed default serialization ID. RevAPI reports this as `java.class.defaultSerializationChanged`, so these entries acknowledge the two expected changes.

api/src/main/java/org/apache/iceberg/types/AssignIds.java
```diff
@@ -56,11 +56,7 @@ public Type struct(Types.StructType struct, Iterable<Type> futures) {
     for (int i = 0; i < length; i += 1) {
       Types.NestedField field = fields.get(i);
       Type type = types.next();
-      if (field.isOptional()) {
-        newFields.add(Types.NestedField.optional(newIds.get(i), field.name(), type, field.doc()));
-      } else {
-        newFields.add(Types.NestedField.required(newIds.get(i), field.name(), type, field.doc()));
-      }
+      newFields.add(Types.NestedField.from(field).withId(newIds.get(i)).ofType(type).build());
```
mkuchenbecker, line 59: Is there no more differentiation or is it implicit with the from?

mkuchenbecker:
1. One comment on a potential behaviour change.
2. Can you verify that "Backport rewrite delete file fix (#10020) to iceberg 1." was a clean cherry pick? What were the commands run and were there conflict before the commit was made. 
Ideally you cherry pick -x to preserve the original commit, and commit as-is with any merge conflicts. Then you update code. That there are updates but no conflicts makes me think the backport wasn't clean. 

A +1 response to this comment to indicate it actually was clean I'll take vs redoing.

api/src/main/java/org/apache/iceberg/types/AssignIds.java
jiang95-dev: `from()` will copy everything from the old field. This is a new method added by @cbb330 to support default values.

jiang95-dev:
> 1. One comment on a potential behaviour change.
> 2. Can you verify that "Backport rewrite delete file fix (#10020) to iceberg 1." was a clean cherry pick? What were the commands run and were there conflict before the commit was made.
>    Ideally you cherry pick -x to preserve the original commit, and commit as-is with any merge conflicts. Then you update code. That there are updates but no conflicts makes me think the backport wasn't clean.
> 
> A +1 response to this comment to indicate it actually was clean I'll take vs redoing.

I just used `git cherry pick -x `to preserve the original commit. Also commented on the `from()` behavior.

mkuchenbecker (approved):

PR #278, opened 2026-09-17
Spark 3.5: retain tracked-column stats for Column-Value Lineage (CVL)

att10:
## What

Enables per-file min/max **bound retention** on the Spark 3.5 / Iceberg 1.5 **batch read** path so downstream Column-Value Lineage (CVL) can compute value bounds for the tracked column(s). Previously the batch scan dropped column stats, so the CVL adapter (li-openhouse) had no bounds to read; it cannot re-add stats after planning, so the retention must be enabled here in the Iceberg Spark integration.

## How

- `SparkSQLProperties` — new session flag `spark.lineage.columnValues.enabled` (default **on**).
- `SparkReadConf` — `columnValueLineageEnabled()` accessor.
- `SparkScanBuilder`:
  - Table-property constants `lineage.columnValues.enabled` (per-table kill switch, default on) and `lineage.columnValues.columns` (default `datepartition`).
  - `lineageStatsColumns(...)` helper: returns empty when disabled by the session flag **or** the per-table kill switch; otherwise resolves the configured columns honoring the scan's `caseSensitive` flag and returns their **canonical** schema names.
  - `buildBatchScan(...)` calls `scan.includeColumnStats(statsColumns)` (requested-columns overload → bounded overhead) inside a **fail-safe** block, so any error skips retention without failing the query.

Default behavior tracks only `datepartition`; overridable per table.

## Scope

Batch query scan only. Incremental / changelog / copy-on-write / aggregate-pushdown paths and write-path CVL are unchanged. No changes to li-openhouse or the engine fork (separate repos).

## Testing

New `TestColumnValueLineageStats` (7 cases): default retention, non-tracked columns dropped, session flag off, per-table kill switch, columns override, missing-column ignored, and case-insensitive resolution. All pass on the `iceberg-spark-3.5_2.12` module.


spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -461,6 +469,15 @@ private Scan buildBatchScan(Long snapshotId, Long asOfTimestamp, String branch,
 
     scan = configureSplitPlanning(scan);
 
+    try {
+      List<String> statsColumns = lineageStatsColumns(expectedSchema);
+      if (!statsColumns.isEmpty()) {
+        scan = scan.includeColumnStats(statsColumns);
```
yangyangv2, line 475: the collected statsColumns are case sensitive (from the config), would the is the scanner case sensitive or insensitive? 

e.g. config = Datepartition, column = datepartition, would they match?

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -461,6 +469,15 @@ private Scan buildBatchScan(Long snapshotId, Long asOfTimestamp, String branch,
 
     scan = configureSplitPlanning(scan);
 
+    try {
+      List<String> statsColumns = lineageStatsColumns(expectedSchema);
```
yangyangv2, line 473: Would it be worth of moving this configuration out, e.g. as a static across the application lifecycle?

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
```diff
@@ -471,6 +488,50 @@ private Scan buildBatchScan(Long snapshotId, Long asOfTimestamp, String branch,
...
   }
 
+  /**
+   * Resolves the columns whose file-level stats (min/max bounds) must be retained on planned tasks
+   * so downstream Column-Value Lineage (CVL) can compute value bounds. Returns an empty list when
+   * CVL is disabled by the session flag or the per-table kill switch, keeping the read path
+   * unaffected.
+   */
+  private List<String> lineageStatsColumns(Schema expectedSchema) {
+    if (!readConf.columnValueLineageEnabled()) {
```
yangyangv2, line 498: spark can also pass session config to disable the column value lineage connection, can you check if it can be also checked here? 

e.g. ```spark.lineage.columnValues.enabled=false```

https://github.com/linkedin-multiproduct/spark-grid-template/pull/70/changes#diff-c42ae28bed8f629bb69576fb75ee2b9dfa378ff6055febffcc79d66bf04a536aL609

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
att10: They should match when `caseSensitive=false` which I believe is the spark default. By the time we call `includeColumnStats(statsColumns)`, the names are already canonicalized. In `lineageStatsColumns` we resolve each configured name honoring the scan's `caseSensitive` flag:
```
Types.NestedField field =
    caseSensitive ? expectedSchema.findField(name)
                  : expectedSchema.caseInsensitiveFindField(name);
if (field != null) {
  statsColumns.add(field.name());   // canonical, exact-case schema name
}
```
- `caseSensitive=false` (Spark default): config Datepartition → `caseInsensitiveFindField` → resolves to datepartition, and we pass the canonical `field.name()` downstream, so the later case-sensitive `findField` in `includeColumnStats` matches.
- `caseSensitive=true`: matching is exact, so Datepartition ≠ datepartition → `findField` returns `null` → the null guard skips it (no NPE, no bounds).

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
att10: - `lineage.columnValues.enabled` and `lineage.columnValues.columns` are per-table and read from `table.properties()`, so two tables in the same app can have different settings.
- `spark.lineage.columnValues.enabled` is per-session (Spark conf), so it can change between sessions/queries.

So I think they have to be resolved per scan? Let me know if I'm misunderstanding your ask though

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
att10: Yup `columnValueLineageEnabled()` checks the value of `spark.lineage.columnValues.enabled`

spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java
yangyangv2: you're right about the table properties. `lineage.columnValues.columns`, every scan needs to resolve it.

The session properties `spark.lineage.columnValues.enabled` can be at the session level instead of scan-level.

jiang95-dev (changes requested):
Lineage is a concept in the catalog not table format. Including lineage implementation at iceberg level is not right.

Let's redesign like this: In iceberg repo we just accept columns as input and build the Scan using the columns. Then in li-openhouse repo we build the lineage adaptor, and pass these lineage specific columns to the builder. We should also be careful to call iceberg-1.2 and iceberg-1.5 in a compatible way.

yangyangv2 (approved):

jiang95-dev (approved):

